PR #21, opened 2019-05-27
feat(API): implement the keys_create and keys_balance function

### LINKED ISSUE #15: Implementation of '$ safe key create' subcommand
Creation of keys linked to a Wallet and also anonymous Keys.

### LINKED ISSUE #16: Special command to allocate test-safecoins into a Key (only for mock/testing)

bochaco:
Resolves #15 
Resolves #16 

This PR implements the following commands:
- keys create --anon --preload \<amount\> --test-coins
- keys create --anon --preload \<amount\> --from \<Key xorname\>
- keys balance --target \<Key xorname\>


Cargo.toml
```diff
@@ -8,36 +8,25 @@ edition = "2018"
 license = "GPL-3.0"
```
dirvine: if we have 
edition = "2018" 
Then it locks in place all the 2018 features (like async soon)

src/lib.rs
```diff
@@ -1 +1,100 @@
+use threshold_crypto::{PublicKey, SecretKey};
 
+pub mod scl_mock;
+use scl_mock::{MockSCL, XorName};
+
+pub struct BlsKeyPair {
+    pub pk: PublicKey,
+    pub sk: SecretKey,
```
dirvine: This should not really be public. The best way is to keep it private and create the sign verify etc. around this. Decent example https://github.com/poanetwork/threshold_crypto/blob/master/examples/basic_pkc.rs

.travis.yml
```diff
@@ -5,7 +5,7 @@ env:
     - PATH=$PATH:$HOME/.cargo/bin
 language: rust
 rust:
-  - 1.32.0
+  - 1.34.2
```
dirvine: you can set this to just `- stable` and it will always be the latest stable. Just saves you updating this one.

src/lib.rs
bochaco: This is for the API to return the key pair generated, e.g. when the users creates a new Key/CoinBalance but asks the API to generate the key pair for it, so the API returns it for the user to then decide what to do with the secret key, or even with the public key if it's gonna publish it somewhere like in a WebID/Wallet/Website...

bochaco: Although, perhaps we can return our own data types instead of the threshold_crypto PublicKey/PrivateKey, perhaps just as array of bytes

Cargo.toml
bochaco: I just moved it to be the last in the section.

PR #26, opened 2019-06-03
feat(Wallet): implementation of lib API and CLI command for creation of Wallets

### LINKED ISSUE #9: Implementation of '$ safe wallet create' subcommand
We need to make sure this is encrypted also as it will contain the secret keys of the CoinBalances

### LINKED ISSUE #10: Implementation of '$ safe wallet balance' subcommand

### LINKED ISSUE #22: Implementation of internal API for MutableData container
We need our API for MD which provides us with the abstractions will need to our CLI use cases, e.g. creating a wallet which contains the links to Keys and their secret keys, eventually we'll expose a set of commands for the user to also create/update and delete an MD form the CLI.

### LINKED ISSUE #25: Special command to allocate test-safecoins into a Key and add it to a Wallet (only for mock/testing)

bochaco:
Resolves #9 
Resolves #25 
Resolves #10 
Resolves #22


joshuef:
I don't think we need / should have `MutableData` commands in the CLI (if we can avoid it). We may need that internally, but whatever commands for manipulating this data, I think it should form part of a `safe data` command if we need it. Or perhaps be flag controlled from `Containers` ?

bochaco:
I was imagining that since MD is effectievely one of the Unpublished native data types, users may want to have access to them in a raw way, if we have CLI for MD exposing a nice key-value data model this becomes the new Redis?

joshuef:
Aye i'm not against that per se. But I think we probably want to keep a lid on the data types. I worry about it getting like the safe_app_nodejs with a bunch of different / quite low level things.

Maybe:
```shell
// appendable / unsequenced / unpublished by default 
$ safe data create --mutable --sequenced`
// update or insert?
$ safe data update <key> <value> [<version if needed>]`
//error if published
$ safe data remove <key or xorurl> [<version if needed>]`
```

joshuef:
also worth considering how this relates to / overlaps with `safe container`

bochaco:
> I worry about it getting like the safe_app_nodejs with a bunch of different / quite low level things.
Yes, me too, and I'd say it'll depend on how the SCL API end up being, personally, if they ar enot like a key-value nice thing, I'd prefer a higher abstraction

PR #27, opened 2019-06-05
feat(xor-url): add support for XOR-URL using CID

bochaco:
Also allow base32 (default) and base32z encoding to be chosen by the user, e.g.:
```
$ safe keys create --test-coins --xor-url base32z
Note that the Key to be created will be preloaded with **test coins** rather than real coins
You must pass a preload amount with test-coins, 1000.111 will be added
New Key created at XOR-URL: "safe://hbkwmnbefby6w8dzqkenwh9imryt4nnck47uy7dggzjqnyozgwgdpqamz4"
Key pair generated: pk="a0a107a871ddca40a9cfd564047421315aecc1d198d74b84085cd430daec2efadd29f1fdf81ac16e1671980e7b3232f4", sk="f90bcb6e386fc7c3dc3e8a40f2dd1a677670d975be30fb05429bd4cb1101ba3a"
```

If the provided base is not supported is uses the default:
```
$ cargo run -- keys create --test-coins --xorurl base64
Note that the Key to be created will be preloaded with **test coins** rather than real coins
You must pass a preload amount with test-coins, 1000.111 will be added
Base encoding 'base64' not supported for XOR-URL. Using default 'base32'.
New Key created at XOR-URL: "safe://bbkulcblpuelwirijnieunkbhqmsqcrv5cagtfaovcwgsdsue5dtecbzks"
Key pair generated: pk="adf422ec88a12d4128d504f064a028d7a201a6503aa2b1a439509d1cc820e552cb657abdaaf89211fd3763be8c1d0267", sk="ffb186c39db8d94e9ccf09791255890bb9ae246742e6bb1e4c75182419333b48"
```

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -42,14 +42,14 @@ struct CmdArgs {
     dry: bool,
     /// Base encoding to be used for XOR-URLs generated. Currently supported: base32 (default) and base32z
     #[structopt(long = "xorurl", raw(global = "true"))]
-    xor_url_base: Option<String>,
+    xorurl_base: Option<String>,
 }
 
 pub fn run() -> Result<(), String> {
     // Let's first get all the arguments passed in
     let args = CmdArgs::from_args();
 
-    let mut safe = Safe::new(args.xor_url_base.clone().unwrap_or("".to_string()));
+    let mut safe = Safe::new(args.xorurl_base.clone().unwrap_or("".to_string()));
 
     debug!("Processing command: {:?}", args);
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -40,13 +40,16 @@ struct CmdArgs {
     /// Dry run of command. No data will be written. No coins spent.
     #[structopt(long = "dry-run", raw(global = "true"))]
     dry: bool,
+    /// Base encoding to be used for XOR-URLs generated. Currently supported: base32 (default) and base32z
+    #[structopt(long = "xorurl", raw(global = "true"))]
+    xor_url_base: Option<String>,
 }
 
 pub fn run() -> Result<(), String> {
     // Let's first get all the arguments passed in
     let args = CmdArgs::from_args();
 
-    let mut safe = Safe::new();
+    let mut safe = Safe::new(args.xor_url_base.clone().unwrap_or("".to_string()));
 
     debug!("Processing command: {:?}", args);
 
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -40,13 +40,16 @@ struct CmdArgs {
     /// Dry run of command. No data will be written. No coins spent.
     #[structopt(long = "dry-run", raw(global = "true"))]
     dry: bool,
+    /// Base encoding to be used for XOR-URLs generated. Currently supported: base32 (default) and base32z
+    #[structopt(long = "xorurl", raw(global = "true"))]
+    xor_url_base: Option<String>,
```
joshuef: @bochaco since you're touching a few things here. Maybe this is a good time to align on `xor_url` or `xorurl` for naming. I think `xorurl` is maybe just easier (i know how you hate hyphens, so underscores must be _painful_ for you :P ) ? But aye, using one across the board will make things cleaner.

src/lib_helpers.rs
```diff
@@ -107,3 +86,75 @@ pub fn sk_from_hex(hex_str: &str) -> SecretKey {
     let sk_bytes = parse_hex(&hex_str);
     bincode::deserialize(&sk_bytes).expect("Failed to deserialize provided secret key")
 }
+
+pub fn xorname_to_xorurl(xorname: &XorName, base: &String) -> String {
```
joshuef: should we start using the rust documentation test system so we're generating examples as we go/test?

src/cli.rs
bochaco: hahaha, yes I don't like them in the user facing commands, I don't mind either way, here in the code I was just following some type of convention where you separate words in var names, if it was JS you'd call it `xorUrl` and in here it translates to `xor_url`, and `xorName` would need to be `xor_name`. But for this particular case I don't mind, in fact if we are trying to stablish the name XOR-URL perhaps xorurl is just fine. I'll update this PR if we are on the same page @joshuef ?

joshuef: aha, yeh, i dont mind either way. hyphen or no.

just one way (as there's currently both in the code)

src/lib_helpers.rs
bochaco: ah yes, we should!

bochaco: although that particular function may not be exposed, or may yes we do expose it as part of a XOR-URL utility lib/API?

Issue #5, opened 2019-05-22
User guide


bochaco:
It can be simply all the info from the RFC and any additional examples as needed, plus instructions to build the project and run it.

calumcraig:
User guide has been updated with more detailed instructions / information.

Moving this to done.

PR #28, opened 2019-06-06
Feat wallet transfer

### LINKED ISSUE #11: Implementation of '$ safe wallet transfer' subcommand
Support to transfer `--from` & `--to` with Wallets/Ledgers, Keys and BLS-pk

joshuef:
Todo: 
- [x] wallet integration tests

Fixes #11 and tracking #5

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@ use scl_mock::MockSCL;
 use serde::{Deserialize, Serialize};
 use threshold_crypto::SecretKey;
 use unwrap::unwrap;
+use uuid::Uuid;
 
 static WALLET_TYPE_TAG: u64 = 10000;
 
@@ -32,11 +33,10 @@ pub struct BlsKeyPair {
 }
 
 // Struct which is serialised and stored in Wallet MD for linking to a spendable balance (Key)
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Debug)]
 struct WalletSpendableBalance {
     xorurl: XorUrl,
     sk: String,
-    default: bool,
 }
 
 pub struct Safe {
@@ -133,7 +133,7 @@ impl Safe {
     pub fn keys_fetch_pk(&self, xorurl: &XorUrl, sk: &str) -> String {
         let secret_key: SecretKey = sk_from_hex(sk);
         let xorname = xorurl_to_xorname(xorurl);
-        let public_key = self.safe_app.keys_fetch_pk(&xorname, &secret_key);
+		let public_key = self.safe_app.keys_fetch_pk(&xorname, &secret_key);
         pk_to_hex(&public_key)
     }
 
@@ -155,7 +155,6 @@ impl Safe {
         let value = WalletSpendableBalance {
             xorurl: key_xorurl.clone(),
             sk: key_pair.sk.clone(),
-            default,
         };
         let serialised_value = unwrap!(serde_json::to_string(&value));
         // FIXME: it should return error if the name already exists
@@ -167,6 +166,16 @@ impl Safe {
             &k,
             &serialised_value.into_bytes(),
         );
+
+        if default {
+            // add the _default key
+            self.safe_app.mutable_data_insert(
+                &wallet_xorname,
+                WALLET_TYPE_TAG,
+                &"_default".to_string().into_bytes(),
+                &k,
+            );
+        }
     }
 
     // Check the total balance of a Wallet found at a given XOR-URL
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -130,10 +130,9 @@ impl Safe {
     }
 
     // Fetch Key's pk from the network from a given XOR-URL
-    pub fn keys_fetch_pk(&self, xorurl: &XorUrl, sk: &str) -> String {
-        let secret_key: SecretKey = sk_from_hex(sk);
+    pub fn keys_fetch_pk(&self, xorurl: &XorUrl) -> String {
         let xorname = xorurl_to_xorname(xorurl);
-        let public_key = self.safe_app.keys_fetch_pk(&xorname, &secret_key);
+		let public_key = self.safe_app.keys_fetch_pk(&xorname);
```
bochaco: I see this wrong format in a few places, perhaps you are not running `cargo fmt`?

src/lib.rs
```diff
@@ -178,17 +186,144 @@ impl Safe {
...
+        let xorname = xorurl_to_xorname(&wallet_xorurl);
+        let mut default_key: String = "".to_string();
+
+        if let Some(default) =
+            unwrap!(self
+                .safe_app
+                .mutable_data_get_key("_default", &xorname, WALLET_TYPE_TAG))
+        {
+            default_key = String::from_utf8_lossy(&default).to_string();
+            println!(
```
bochaco: If this is meant to be for the pretty output of the CLI it should then be moved out of here, and here it should be a log info message instead.

src/lib.rs
```diff
@@ -178,17 +186,144 @@ impl Safe {
...
+    ///
+    /// Using established Wallet and SpendableBalances you can send safecoins between Wallets.
+    ///
+    /// ## Example
+    /// ```
+    /// # use safe_cli::Safe;
+    /// # use unwrap::unwrap;
+    /// let mut safe = Safe::new("base32".to_string());
+    /// let sk = String::from("391987fd429b4718a59b165b5799eaae2e56c697eb94670de8886f8fb7387058");
+    /// let wallet_xorurl = safe.wallet_create();
```
bochaco: I guess we then rename things like `wallet_xor_url` to `wallet_xorurl` ?

src/lib.rs
```diff
@@ -178,17 +186,144 @@ impl Safe {
...
+            .keys_fetch_pk(&xorurl_to_xorname(&from_wallet_balance.xorurl));
+
+        let to_pk = self
+            .safe_app
+            .keys_fetch_pk(&xorurl_to_xorname(&to_wallet_balance.xorurl));
+
+        let from_sk = sk_from_hex(&from_wallet_balance.sk);
+        let tx_id = Uuid::new_v4();
+
+        // TODO: what doe this return? tx_id?
```
bochaco: I think we can **eventually** return `Result<tx_id, ErrorType>` ? or even `Result<(), ErrorType>`

src/lib.rs
```diff
@@ -178,17 +186,144 @@ impl Safe {
...
+        self.safe_app
+            .safecoin_transfer(&from_pk, &from_sk, &to_pk, &tx_id, amount);
+
+        println!("Success. TX_ID: {:?}", &tx_id);
```
bochaco: This again is at most a log info message. the println should be moved to the CLI

src/lib.rs
```diff
@@ -338,7 +473,7 @@ fn test_keys_fetch_pk_test_coins() {
     let mut safe = Safe::new("base32".to_string());
     let (xorurl, key_pair) = safe.keys_create_preload_test_coins("23.22".to_string(), None);
     let key_pair_unwrapped = unwrap!(key_pair);
-    let pk = safe.keys_fetch_pk(&xorurl, &key_pair_unwrapped.sk);
+    let pk = safe.keys_fetch_pk(&xorurl);
```
bochaco: Just to know why you removed this arg from key_fetch_pk, we now don't care but whenever we use the SCL API fr sure will need to pass the sk as it will be an OnwedGet, or least that's my understanding. So I put it to force us to have the right info passed around even that still it's not being used. Not a big deal, but making sure we are on the same page

src/subcommands/keys.rs
```diff
@@ -75,11 +85,11 @@ pub fn create_new_key(
             safe.keys_create(from, preload, pk)
         } else {*/
         warn!("Note that the Key to be created will be preloaded with **test coins** rather than real coins");
-        println!("Note that the Key to be created will be preloaded with **test coins** rather than real coins");
+        // println!("Note that the Key to be created will be preloaded with **test coins** rather than real coins");
         let amount = preload.unwrap_or("1000.111".to_string());
 
         if amount == "1000.111" {
-            eprintln!("You must pass a preload amount with test-coins, 1000.111 will be added");
+            warn!("You must pass a preload amount with test-coins, 1000.111 will be added by default.");
```
bochaco: nice! i was thinking the same last night about saying that it's a default value

src/lib.rs
joshuef: Yeh right now im going for `Ok` from this func. But i was more wondering what makes most sense coming back from the raw SCL api

src/lib.rs
joshuef: aha, right I was wondering as it wasn't used, and some clippy or some such was nagging me. I'll re add 👍

src/lib.rs
joshuef: Hmm, we can't need a secret to get the PK can we? If so it's not public?

src/lib.rs
joshuef: Hmm, but we shouldnt need SK to get PK should we?

bochaco: It depends what exactly such an API/RPC is retrieving, if it's retrieving only the pk then maybe not, but if it's also retrieving the balance then you want to have it with OwnedGet presumably.

src/subcommands/wallet.rs
```diff
@@ -97,37 +117,59 @@ pub fn wallet_commander(cmd: Option<WalletSubCommands>, safe: &mut Safe) -> Resu
...
+                                linked_key
+                            ),
+                            "Invalid input",
+                        );
+                    }
+
+                    let pk = safe.keys_fetch_pk(&linked_key);
+                    if pretty {
+                        println!(
+							"Spendable balance added with name '{}' in wallet located at XOR-URL \"{}\"",
```
bochaco: You still seem to be having some formatting issues?

PR #108, opened 2019-06-27
feat(target): use CARGO_TARGET_DIR for integration tests

joshuef:
Enables use of env var for setting build/test output

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -1,8 +1,17 @@
-pub fn get_bin_location() -> &'static str {
-    let mut location = "./target/release/safe_cli";
-    if cfg!(debug_assertions) {
-        location = "./target/debug/safe_cli";
-    }
+use std::env;
+
+pub fn get_bin_location() -> String {
+    let location = match env::var("CARGO_TARGET_DIR") {
+        Ok(location) => location,
+        Err(e) => {
+            let mut location = "./target/release/safe_cli";
+            if cfg!(debug_assertions) {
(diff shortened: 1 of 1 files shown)


joshuef:
FYI @jacderida

tests/common/mod.rs
```diff
@@ -1,8 +1,17 @@
-pub fn get_bin_location() -> &'static str {
-    let mut location = "./target/release/safe_cli";
-    if cfg!(debug_assertions) {
-        location = "./target/debug/safe_cli";
-    }
+use std::env;
+
+pub fn get_bin_location() -> String {
+    let location = match env::var("CARGO_TARGET_DIR") {
```
jacderida: Sorry I may be misunderstanding, but would we not still need the `release/safe_cli` appended to the `CARGO_TARGET_DIR` location?

Sorry Rust isn't my primary language, so could be completely wrong!

jacderida: Yeah sorry, it doesn't quite work. On my branch I've changed it to:
```
use std::env;

pub fn get_bin_location() -> String {
    let target_dir = match env::var("CARGO_TARGET_DIR") {
        Ok(target_dir) => target_dir,
        Err(e) => "./target".to_string()
    };
    if cfg!(debug_assertions) {
        return format!("{}{}", target_dir, "/debug/safe_cli");
    }
    format!("{}{}", target_dir, "/release/safe_cli")
}
```

That seems to work ok.

PR #113, opened 2019-07-03
Jenkins Build and Release

jacderida:
Hi Josh/Gabriel

Jenkins now provides a build and deployment process for all platforms. Travis has been stripped down to only running Linux.

## Build Process

* In parallel:
  - Build Linux in container
  - Build Windows
  - Build macOS
  - Upload build artifacts to S3
  
For Linux, the container definition is in Dockerfile.build. The use of a .build extension is to differentiate between a container for distributing the application, which we may add at some point in the future.

At this point the time for the Linux build is quite slow. I've got some optimisations I can do for this, but those are mostly infrastructure changes and I'd like to get this merged first.

## Deployment Process

The deployment process will only occur on the master branch, i.e. after a PR is merged.

* Clone the code with the tags fetched (not done by default)
* Retrieve build artifacts from S3
* If the commit message begins with the string "Version change":
  - Get the version number
  - Package the artifacts with the version number in the file name
  - Create a tag for this version
  - Publish a GitHub release with the artifacts attached
* Otherwise:
  - Package the artifacts with commit hash in the file name
  - Upload the artifacts to the safe-cli S3 bucket
  
[Here](https://jenkins.maidsafe.net/blue/organizations/jenkins/pipeline-safe_cli/detail/PR-113/15/pipeline/) is a build with a deployment to S3.
[Here](https://jenkins.maidsafe.net/blue/organizations/jenkins/pipeline-safe_cli/detail/PR-113/20/pipeline/) is a build with a version change commit that published to [GitHub](https://github.com/maidsafe/safe-cli/releases/tag/0.0.2). (This commit was then reverted, and I'll remove the release from GitHub when this is merged.)

The latest build does not do a deployment as I put in the conditional on the master branch. That's expected behaviour.

Cheers,

Chris

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,18 +3,12 @@ env:
     - RUST_BACKTRACE=1
     # - RUSTFLAGS="-C opt-level=2 -C codegen-units=8"
     - PATH=$PATH:$HOME/.cargo/bin
+    - YARN_GPG=no # otherwise this starts gpg-agent that never exits
+os:
+  - linux
 language: rust
 rust:
   - stable
-
-matrix:
-  include:
-    - os: linux
-    - os: osx
-    - os: windows
-      env:
-        - YARN_GPG=no # otherwise this starts gpg-agent that never exits
-
 sudo: false
 cache:
   # Double the default timeout.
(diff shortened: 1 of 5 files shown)


bochaco:
This looks really nice @jacderida ! Jut a heads up for you to think of, eventually we''ll need another type of deploy which is the crate on creates.io, this project has to deliverables, on is the CLI application which is what you are building and deploying, but it also has a lib with API for SAFE Rust developers, which means we will publish a crate on crates.io for them to incude it in the Rust applications and use the API. 

The CLI application makes use of this API, it's just another application like those that devs will create using this crate's API.

Issue #126, opened 2019-07-04
Files put: Differentiate default root via trailing `/`


joshuef:
Similar to `cp`.

Without a trailing slash, the folder name is used as the root, unless another is specified.

With a slash, the folder contents are set as the root.

ie:

`safe put somewhere`

results in keys:

`somewhere/index.html` in the filesmap.

and

`safe put somewhere/`

results in keys:

`index.html` in the filesmap.

calumcraig:
Without trailing /:

```
$ ./safe files put upload-stuff -r
FilesContainer created at: "safe://hbys4q7crof8sju8rspck5qcqxyg9ujbotc5ufc49w7k8gdj1bjohf4j5w"
+  upload-stuff/image.png                           safe://hox1cn1oeeygux97hxbx9ubmk8k8jfscni96raiogwb96d8g4gnbgw 
+  upload-stuff/newfile                             safe://hoqn7brhocuj5kqgs33gr9ggyzrj1hndog8f8bk9p4csy1iyt3z7kn 
+  upload-stuff/subfolder1/subfolder2/index.html    safe://hoqd8xnmacwg9qxfhtycmtnpnyqaauwfomzm64k7j74maeaxr37gbc 
+  upload-stuff/subfolder1/subfolder2/picture.jpeg  safe://hoxfnxbkfkn4oy9gyzhnrgasr1sb1znpstgwyotfemyw3n4dc6e3fa 
+  upload-stuff/subfolder1/testfile4.index          safe://hoq96qzd3zfocxmm5e8qjncxx5ez6ayareqkqfbg51wccpnthmib18 
+  upload-stuff/testfile.txt                        safe://hoq9sr6uxammrm4cgi7osfcq5d5swrn6sbdc936gzye9gncagmjznn 
+  upload-stuff/testfile1.txt                       safe://hox66i3xeb1g1wmo55ssjbadugh97o9xow5uejk5n6i79acycyiabu 
+  upload-stuff/testfile2.txt                       safe://hoq3bakodjcnojgqu1mjy8zt339gzas5u8nq8ybjuyue7pzy8rg1eq
```
```
$ ./safe cat safe://hbys4q7crof8sju8rspck5qcqxyg9ujbotc5ufc49w7k8gdj1bjohf4j5w
Files of FilesContainer (version 1) at "safe://hbys4q7crof8sju8rspck5qcqxyg9ujbotc5ufc49w7k8gdj1bjohf4j5w":
[...]

PR #130, opened 2019-07-05
fix(Files): Trailing slash for files commands now indicates to use fo…

joshuef:
#126


src/subcommands/files.rs
```diff
@@ -56,6 +58,28 @@ pub enum FilesSubCommands {
     },
 }
 
+fn get_the_real_root( location : &str, root : Option<String> ) -> Option<String> {
```
bochaco: I'd move this to be part of the lib, so it's also available for other apps

Issue #119, opened 2019-07-04
providing invalid sk when creating wallet causes cli crash when querying wallet balance


calumcraig:
**Steps to reproduce:**
1. `safe wallet create <valid_public_key>`
2. When prompted for the corresponding secret key enter an invalid string. Eg:
```
$ ./safe wallet create safe://hodqfq1zkktgqcahp91qmbj5xfewhsmpe6ywucchb9spdwrztwww7g
Enter secret key corresponding to public key at "safe://hodqfq1zkktgqcahp91qmbj5xfewhsmpe6ywucchb9spdwrztwww7g": bdfvrf
Wallet created at: "safe://hbyewc61at9r3xkftkgqw3z9wgynn4776dbs437d7qh8r8e3txa1dn6jao"
```
3. `safe wallet balance <wallet_xorurl_provided_above>`

**Expected:**
Invalid secret key is detected and wallet creation prevented, _and / or_ error at wallet balance stage is handled gracefully.

**Actual:**
safe_cli crash:
```
Well, this is embarrassing.

safe_cli had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "/tmp/report-969f0ed0-9b9b-4a45-8a50-e7a66207261e.toml". Submit an issue or email with the subject of "safe_cli Crash Report" and include the report as an attachment.

- Authors: bochaco <gabrielviganotti@gmail.com>, Josh Wilson <joshuef@gmail.com>

We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
```

calumcraig:
#118 was tested to ensure that `wallet create` and `wallet insert` command will no longer accept invalid `sk` for the keys used to pay for the `wallet` transactions.

The scenario described in this bug is therefore no longer a valid one.

QA verified (see [here](https://github.com/maidsafe/safe-cli/issues/118#issuecomment-510807185) for test results), moving this to done.

PR #140, opened 2019-07-10
fix(Wallet): Validate SK and PK upon wallet creation.

### LINKED ISSUE #118: wallet create / insert: validate sk
When the user provides a valid public key for `wallet create` or `wallet insert` they are prompted for the corresponding secret key. Eg:
```
$ ./safe wallet insert safe://hodniyu7oqcyks1e4g8fte9gapex9puot8hrtfjdaxidh1yzw9wu96 safe://hbymngmtpyss8aftmfscdkeow5ob5kkz7frokfi7x8htdgri37z89e6jao safe://hbyma648f5xprwmmy44xsk4si16dsx9y46ft1a8bxfnhu36enbc1cyyjao
Enter secret key corresponding to public key at "safe://hodniyu7oqcyks1e4g8fte9gapex9puot8hrtfjdaxidh1yzw9wu96":
```
However, it is possible to input anything at this stage and the safe_cli "accepts" this and returns a success message:
```
Spendable balance inserted with name 'safe://hodniyu7oqcyks1e4g8fte9gapex9puot8hrtfjdaxidh1yzw9wu96' in Wallet located at "safe://hbyma648f5xprwmmy44xsk4si16dsx9y46ft1a8bxfnhu36enbc1cyyjao"
```

This input should be validated to ensure the secret key is the correct, corresponding key to the original public key provided.

Two initial suggestions:
1. Validate the number of characters of the `sk` provided.
2. Check the `sk` against the `pk` provided.

joshuef:
+ adds validation checks
+ adds integration tests

fixes #118

tracking #119


src/api/wallet.rs
```diff
@@ -55,6 +55,12 @@ impl Safe {
             xorurl: key_xorurl.to_string(),
             sk: key_pair.sk.clone(),
         };
+
+        match validate_key_pair(key_pair.clone()) {
```
bochaco: @joshuef I think this can be replaced with a single line statement: `validate_key_pair(key_pair.clone())?;`

src/api/keys.rs
```diff
@@ -12,14 +12,42 @@ use super::{Safe, XorUrl, XorUrlEncoder};
...
-#[derive(Clone)]
+#[derive(Clone, Validate)]
 pub struct BlsKeyPair {
+    #[validate(length(equal = "96"))]
     pub pk: String,
+    #[validate(length(equal = "64"))]
     pub sk: String,
 }
 
+pub fn validate_key_pair(key_pair: BlsKeyPair) -> Result<(), String> {
```
bochaco: Also, perhaps you can have key_pair to be passed as ref so no need to clone it?: `pub fn validate_key_pair(key_pair: &BlsKeyPair)`

PR #156, opened 2019-07-12
feat(keys): integration with SCL CoinBalance API

### LINKED ISSUE #23: Wallet cmd: adapt the internal API to use the SCL API instad of SCL-mock

### LINKED ISSUE #82: `test_wallet_transfer_no_default` unit test failing with feature `mock-network`
When running unit tests against SCL mock-network one test (`test_wallet_transfer_no_default`) is failing due to a bug in SCL. We depend on a fix for issue https://github.com/maidsafe/safe_client_libs/issues/842

### LINKED ISSUE #141: Setup mock-network tests
With #85 integrated for files / data. We should be able to get the `mock-network` tests to run, at least at the integration level.
(the vault is painfully slow, so I'd say a full test suite run is out of scope for now).

### LINKED ISSUE #158: passing an invalid xorurl to safe cat or files sync causes safe-cli to crash
The xorurl passed `safe://hbywphoqhbkj1kfdmgu8d9xpabc1e4twfpru5b5hjg6azaw6bgyfdooj5w` is the correct length of characters so passes that validation (if xorurl is too long or short validation halts at input, before the safe-cli attempts to process it) but returns a crash:

```
> ./safe cat safe://hbywphoqhbkj1kfdmgu8d9xpabc1e4twfpru5b5hjg6azaw6bgyfdooj5w
Well, this is embarrassing.

safe_cli had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "/tmp/report-6675b578-658d-46ac-a9f7-c4722323f9b0.toml". Submit an issue or email with the subject of "safe_cli Crash Report" and include the report as an attachment.

- Authors: bochaco <gabrielviganotti@gmail.com>, Josh Wilson <joshuef@gmail.com>

We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
```

bochaco:
Resolves #23 
Resolves #82 
Resolves #141 
Resolves #158 

This PR also upgrades to latest code from SCL/safe-nd `stable` branches.

### DIFF
--- a/src/api/safe_client_libs.rs
+++ b/src/api/safe_client_libs.rs
@@ -13,9 +13,8 @@ use futures::future::Future;
 use log::{debug, info, warn};
 use rand::rngs::OsRng;
 use rand_core::RngCore;
-use safe_app::AppError::CoreError;
-use safe_app::{run, App};
-use safe_core::CoreError as SafeCoreError;
+use safe_app::{run, App, AppError::CoreError};
+use safe_core::{client::wallet_transfer_coins, CoreError as SafeCoreError};
 use std::str::FromStr;
 
 #[cfg(not(feature = "fake-auth"))]
@@ -33,8 +32,7 @@ use safe_nd::{
 pub use threshold_crypto::{PublicKey, SecretKey};
 
 use std::collections::BTreeMap;
-use unwrap::unwrap;
-use uuid::Uuid;
+use unwrap::unwrap; // TODO: remove all unwraps from this file
 
 const APP_NOT_CONNECTED: &str = "Application is not connected to the network";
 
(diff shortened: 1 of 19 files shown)


src/api/safe_client_libs.rs
```diff
@@ -71,119 +69,144 @@ impl SafeApp {
...
-        // Ok(balance.to_string())
-
-        // self.scl_mock.get_balance_from_xorname(xorname, sk)
-        Ok("0".to_string())
+    pub fn get_balance_from_sk(&self, sk: SecretKey) -> ResultReturn<String> {
+        let safe_app: &App = self.get_safe_app()?;
+        let coins_amount = run(safe_app, move |client, _app_context| {
+            client
+                .get_balance(Some(&sk))
+                .map_err(|e| CoreError(SafeCoreError::Unexpected(format!("{:?}", e))))
```
joshuef: Do we need this line? It should just be returning the error to the external `map_err` anyway, or?

bochaco: We shouldn't need it @joshuef but `get_balance` seems to be returning an `safe_app::errors::AppError` which is not what `run` expects, so it seems we can mention this to SCL folks for enhancement.

PR #171, opened 2019-07-24
Run Clippy in Parallel

jacderida:
Runs clippy in parallel only on Linux.

So it turns out this doesn't save a huge amount of time, but it seems to stop the Windows machine running out of memory, so I think we should just keep it this way.

It's mainly the Windows build that's really slow and also compiling safe auth is adding in a lot of time, but we can deal with that at another time.


jacderida:
This branch was merged into the self update branch anyway, so I'm just going to close this.

PR #184, opened 2019-07-31
feat(cli): introduce a --pay-with argument for keys and wallet commands

### LINKED ISSUE #176: Wallet cmd: support for providing source Wallet/Key with optional `--pay-with` arg
Rather than having a `<source>` positional argument in wallet commands for the funds, we shall accept the source wallet/Key as an optional argument passed with `--pay-with`. When this is not provided the CLI will assume it shall use the default wallet linked from the user's account (this will required to be logged in with an account).

bochaco:
This is to choose the payer wallet more explicitly in keys and wallet commands

Resolves #176


src/api/fake_scl.rs
```diff
@@ -52,7 +52,7 @@ impl Drop for SafeAppFake {
     fn drop(&mut self) {
         let serialised = serde_json::to_string(&self.fake_vault)
             .expect("Failed to serialised fake vault data to write on file");
-        debug!("Writing serialised fake vault data = {}", serialised);
+        trace!("Writing serialised fake vault data = {}", serialised);
```
joshuef: i forgot trace existed. I approve of this move :)

tests/cli_keys.rs
```diff
@@ -69,7 +69,7 @@ fn calling_safe_keypair_pretty() {
 #[test]
 fn calling_safe_keys_balance() {
     let (pk_xor, sk) = create_preload_and_get_keys("123");
-
+    println!("AA: {}", pk_xor);
```
joshuef: We should remove this in a later PR.

bochaco: oh yes, I think we have to remove all println! we have out there in tests, I'll do that in next PR.

PR #187, opened 2019-08-01
feat(xorurl): support for decoding the version from URLs and fetching the specified version

### LINKED ISSUE #56: Cat cmd: support for retrieving a specific version of a content, plus user guide updates
The user shall be able to retrieve a specific version of the content either when the URL specifies it, e.g. `safe://<NRS/XOR URL>?v=<version>`, or by providing it as an arg in the CLI, e.g. `$ safe cat safe://<NRS/XOR URL> --version <version>`

bochaco:
Resolves #56


bochaco:
This is all now included in PR #188

PR #188, opened 2019-08-02
feat(urls): support versioned URLs and NRS URLs in all commands and API

### LINKED ISSUE #56: Cat cmd: support for retrieving a specific version of a content, plus user guide updates

### LINKED ISSUE #72: NRS cmd: integration tests, and user guide updates

### LINKED ISSUE #177: Support for NRS URLs across all API and CLI commands
Up until now, all commands and API were using XOR-URLs for locating content, we shall support both type of URLs across existing API and CLI, i.e. both NRS and XOR URLs can be provided by the user when providing the location of any type of content (we shall eventually also support HEX encoded XoR names).

### LINKED ISSUE #185: Update documentation for safe cat -i

### LINKED ISSUE #186: make error messages more meaningful

bochaco:
Resolves #56  
Resolves #177 
Resolves #72 
Resolves #186 
Resolves #185


src/subcommands/nrs.rs
```diff
@@ -35,9 +38,9 @@ pub enum NrsSubCommands {
         /// The safe:// URL to map this to. Usually a FilesContainer for a website
         #[structopt(short = "l", long = "link")]
         link: Option<String>,
-        /// Set the sub name as default for this public name
-        #[structopt(long = "default")]
-        default: bool,
+        /// The default is set but using a hard-link pointing to the final destination rather to the sub name (which is the default behaviour)
```
joshuef: I'm not sure 'hard link' makes sense.

Maybe `permalink` ? (which to me at least implies version) or `versioned-link`? Or perhaps we should `follow-link` or `dereference`? 

Actually, does this option link to a specific version or just the underlying ID?

bochaco: This option set the link to the URL you provided in the command with `--link`, so even after removing the subname, the default for public name will be resoled to that specific link you provided, which can be versioned or not.
So, I don't disagree with finding another name, but perm-link doesn't seem to fit as it's not a perm-link if the user provided a non-versioned URL, `dereference` is probably too complex and meaningless to most of people or non-techy? ...let's find out other options if hard-link to too techy as I think it may be

bochaco: I just pushed a commit renaming it to `--direct`

tests/cli_cat.rs
```diff
@@ -75,3 +75,62 @@ fn calling_safe_cat_xorurl_url_with_version() {
...
+        .success();
+
+    // let's sync with another file so we get a new version, and a different content in the file
+    let xorurl_with_path = format!("{}/test.md", container_xorurl);
+    let mut cmd = Command::cargo_bin(CLI).unwrap();
+    cmd.args(&vec!["files", "sync", ANOTHER_FILE, &xorurl_with_path])
+        .assert()
+        .success();
+
+    // NRS name wasn't not updated (with --updated-nrs) when doing files sync,
```
joshuef: `   // NRS name wasn't not` double negative here

tests/cli_nrs.rs
```diff
@@ -482,3 +455,81 @@ fn calling_safe_nrs_put_and_add_new_subnames_remove_one_and_so_fail_to_retrieve(
         ))
         .failure();
 }
+
+/*
+#[test]
```
joshuef: @bochaco Is this intentionally left commented out? Or this functionality doesn't work yet?

bochaco: I just removed it, it is test I was going to write but I covered the scenario in other integration tests like https://github.com/maidsafe/safe-cli/pull/188/files?file-filters%5B%5D=.rs#diff-eaa57147749d5cbad1615fb742671e0dR80

PR #190, opened 2019-08-05
Build on Linux with Musl

jacderida:
Description to be updated.


jacderida:
After quite a bit of extensive investigation, we've concluded that this won't be possible until Safe Client Libs remove their references to rust_sodium. I can't say at this point how complicated that would be.

PR #200, opened 2019-08-08
 feat(nrs): make NRS resolution to work only with versioned links, unless the linked content is unversionable

bochaco:


calumcraig:
Tested this on latest binary: safe_cli-da7c57d-x86_64-unknown-linux-gnu.tar

Test notes attached rather than posting inline comments.

QA verified, moving to done.
[safe-cli_09082019_nrs_versioning.txt](https://github.com/maidsafe/safe-cli/files/3486931/safe-cli_09082019_nrs_versioning.txt)

PR #243, opened 2019-09-11
feat(files): support adding individual files with files commands

### LINKED ISSUE #90: Files add subcommand for Published FilesContainers

bochaco:
- Resolves #90 
- This PR also makes the `files put` command to upload files as `Raw` content-type when their detected media-type is not supported by the XOR-URL encoding, plus minor enhancements to errors reported by wallet API.
- It adds a missing media-type to the list of supported ones: `audio/mpeg`
- Add a `files_container_add_from_raw` API to add a file to an existing `FilesContainer` from raw bytes


tests/cli_files.rs
```diff
@@ -623,3 +623,82 @@ fn files_sync_and_fetch_without_nrs_update() {
...
+    .unwrap();
+    assert_eq!(synced_file_cat, "hello tests!");
+}
+
+#[test]
+fn calling_safe_files_add_a_url() {
+    let files_container_output = cmd!(
+        get_bin_location(),
+        "files",
+        "put",
```
joshuef: we're testing `put` and not `add` here

tests/cli_files.rs
```diff
@@ -623,3 +623,82 @@ fn files_sync_and_fetch_without_nrs_update() {
     assert_eq!(xorurl, nrsurl);
     assert_eq!(files_map.len(), 5);
 }
+
+#[test]
+fn calling_safe_files_add() {
+    let files_container_output = cmd!(
+        get_bin_location(),
+        "files",
+        "put",
```
joshuef: we're testing `put`, instead of `add` here too

PR #248, opened 2019-09-18
Generate native libs for safe-cli and WIP bindings

ravinderjangra:
**Needs to be updated later**


build.rs
```diff
@@ -0,0 +1,181 @@
...
+        lang.add_const("ulong", "SIGN_PUBLIC_KEY_LEN", sign::PUBLICKEYBYTES);
+        lang.add_const("ulong", "SIGN_SECRET_KEY_LEN", sign::SECRETKEYBYTES);
+        lang.add_const("ulong", "XOR_NAME_LEN", XOR_NAME_LEN);
+        lang.add_const("ulong", "BLS_PUBLIC_KEY_LEN", XOR_NAME_LEN);
+        lang.add_opaque_type("App");
+
+        lang.reset_filter(FilterMode::Blacklist);
+
+        let mut outputs = HashMap::new();
+        bindgen.source_file("src/lib.rs");
```
bochaco: This is where we can tweak it to read the FFI lib from e.g. `/src/ffi/mod.rs`, or something like that?

src/api/errors.rs
```diff
@@ -7,6 +7,9 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use std::fmt;
+use ffi_utils::{ErrorCode, StringError};
+use std::ffi::NulError;
```
bochaco: We shouldn't be adding any FFI dependency from any file which is not in `ffi` folder

src/api/errors.rs
```diff
@@ -64,6 +68,87 @@ fn get_error_info(error: &Error) -> String {
         Error::FilesSystemError(info) => ("FilesSystemError".to_string(), info.to_string()),
         Error::Unexpected(info) => ("Unexpected".to_string(), info.to_string()),
         Error::Unknown(info) => ("Unknown".to_string(), info.to_string()),
+        Error::StringError(info) => ("StringError".to_string(), info.to_string()),
     };
     format!("[Error] {} - {}", error_type, error_msg)
 }
+
+#[allow(missing_docs)]
+mod codes {
```
bochaco: I'm not sure we need these codes in Rust API, perhaps only in FFI API? if so I think this should go in e.g. `ffi/errors.rs` file and not here, this is not needed by any Rust developer using this crate therefore it shouldn't be here.

src/api/keys.rs
```diff
@@ -14,6 +14,8 @@ use super::{Error, ResultReturn, Safe, SafeApp, XorUrl, XorUrlEncoder};
 use rand_core::RngCore;
 use serde::{Deserialize, Serialize};
 use threshold_crypto::SecretKey;
+use crate::ffi::ffi_structs::BlsKeyPair as FfiBlsKeyPair;
```
bochaco: There shouldn't be any ffi dependency from any file which is not part of FFI API

src/api/keys.rs
```diff
@@ -28,6 +30,20 @@ pub struct BlsKeyPair {
     pub sk: String,
 }
 
+impl BlsKeyPair {
+    pub fn into_repr_c(self) -> ResultReturn<FfiBlsKeyPair> {
```
bochaco: This is not needed by Rust devs, and thus it doesn't belong to Rust API, thus I think you need to move this to ffi folder: `fn bls_key_pair_into_repr_c(key_pair: &BlsKeyPair) -> ResultReturn<FfiBlsKeyPair>`

src/api/xorurl.rs
```diff
@@ -14,6 +14,7 @@ use multibase::{decode, encode, Base};
 use safe_nd::{XorName, XOR_NAME_LEN};
 use serde::{Deserialize, Serialize};
 use std::fmt;
+use crate::ffi::ffi_structs::XorUrlEncoder as FfiXorUrlEncoder;
```
bochaco: There cannot be any ffi deps in this file since it's not part of Rust API

src/api/xorurl.rs
```diff
@@ -344,6 +392,30 @@ impl XorUrlEncoder {
             None => Ok(xorurl),
         }
     }
+
+    pub fn into_repr_c(self) -> ResultReturn<FfiXorUrlEncoder> {
```
bochaco: This is not needed by Rust devs, and thus it doesn't belong to Rust API, thus I think you need to move this to ffi folder: `fn xorurl_encoder_into_repr_c(xorurl_encoder: &XorUrlEncoder) -> ResultReturn<FfiXorUrlEncoder>`

src/ffi/content_fetch.rs
```diff
@@ -0,0 +1,38 @@
+use ffi_utils::{catch_unwind_cb, from_c_str, FfiResult, OpaqueCtx, FFI_RESULT_OK};
+use std::os::raw::{c_char, c_void};
+use super::{Safe, ResultReturn};
+use crate::api::fetch::SafeData;
+
+#[no_mangle]
+pub unsafe extern "C" fn safe_fetch(
```
bochaco: Why `safe_fetch` and not `fetch` to make it equal to Rust API?

src/ffi/keys.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::api::{Safe};
+use super::ffi_structs::{BlsKeyPair};
+use ffi_utils::{catch_unwind_cb, from_c_str, FfiResult, OpaqueCtx, FFI_RESULT_OK};
+use super::helpers::{from_c_str_to_string_option, to_c_str};
+use std::os::raw::{c_char, c_void};
+
+const PRELOAD_TESTCOINS_DEFAULT_AMOUNT: &str = "1000.111";
+
+#[no_mangle]
+pub unsafe extern "C" fn generate_new_safe_key_pair(
```
bochaco: How about we use a convention to call them exactly the same as in Rust API and just appending maybe `_new`? it'll be much easier to find them when you know the Rust equivalent one. Consistency is king here

src/ffi/keys.rs
```diff
@@ -0,0 +1,94 @@
...
+        let key_xor_url = to_c_str(xorurl)?;
+        let amount_result = to_c_str(PRELOAD_TESTCOINS_DEFAULT_AMOUNT.to_string())?;
+        let ffi_bls_key_pair = key_pair.unwrap().into_repr_c()?;
+        o_cb(user_data.0, FFI_RESULT_OK, key_xor_url.as_ptr(), &ffi_bls_key_pair, amount_result.as_ptr());
+        Ok(())
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn query_key_balance(
```
bochaco: Why renaming it to something different to the one from Rust API? consistency is king

src/ffi/keys.rs
```diff
@@ -0,0 +1,94 @@
...
+            (*app).keys_balance_from_url(&key_url, &sk)?
+        };
+        let amount_result = to_c_str(current_balance)?;
+        o_cb(user_data.0, FFI_RESULT_OK, amount_result.as_ptr());
+        Ok(())
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn transfer_key_balance(
```
bochaco: Again, why diff from Rust equivalent one?

src/ffi/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+pub mod helpers;
+pub mod ffi_structs;
+
+use ffi_utils::{catch_unwind_cb, from_c_str, FfiResult, OpaqueCtx, FFI_RESULT_OK};
+use std::os::raw::{c_char, c_void};
+use crate::api::{Safe, ResultReturn};
+use helpers::{from_c_str_to_str_option};
+
+#[no_mangle]
+pub unsafe extern "C" fn safe_connect(
```
bochaco: how about just `connect` for consistency

src/ffi/nrs.rs
```diff
@@ -0,0 +1,85 @@
+use crate::api::{ResultReturn};
+use crate::api::nrs_map::{NrsMap};
+use crate::api::{Safe};
+use super::helpers::{to_c_str};
+use ffi_utils::{catch_unwind_cb, from_c_str, FfiResult, OpaqueCtx, FFI_RESULT_OK};
+use std::os::raw::{c_char, c_void};
+
+#[no_mangle]
+pub unsafe extern "C" fn create_new_public_name(
```
bochaco: Consistency is king, let's keep it to be the same names as the Rust versions

src/ffi/nrs.rs
```diff
@@ -0,0 +1,85 @@
...
+        let (nrs_map_container_xorurl, _processed_entries, nrs_map) =
+            (*app).nrs_map_container_create(&nrs_str, &link_str, set_default, direct_link, dry_run)?;
+        let xor_url_string = to_c_str(nrs_map_container_xorurl)?;
+        o_cb(user_data.0, FFI_RESULT_OK, &nrs_map, xor_url_string.as_ptr());
+        Ok(())
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn add_update_sub_name(
```
bochaco: Consistency is king, let's keep it to be the same names as the Rust versions

src/ffi/nrs.rs
```diff
@@ -0,0 +1,85 @@
...
+        let (version, xorurl, _processed_entries, nrs_map) =
+            (*app).nrs_map_container_add(&name_str, &link_str, set_default, direct_link, dry_run)?;
+        let xor_url_string = to_c_str(xorurl)?;
+        o_cb(user_data.0, FFI_RESULT_OK, &nrs_map, xor_url_string.as_ptr(), version);
+        Ok(())
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn remove_sub_name(
```
bochaco: Consistency is king, let's keep it to be the same names as the Rust versions

src/ffi/wallet.rs
```diff
@@ -0,0 +1,150 @@
+use crate::api::{ResultReturn};
+use crate::api::{Safe};
+use super::helpers::{from_c_str_to_string_option, to_c_str};
+use ffi_utils::{catch_unwind_cb, from_c_str, FfiResult, OpaqueCtx, FFI_RESULT_OK};
+use std::os::raw::{c_char, c_void};
+
+const PRELOAD_TESTCOINS_DEFAULT_AMOUNT: &str = "1000.111";
+
+#[no_mangle]
+pub unsafe extern "C" fn create_new_wallet(
```
bochaco: Consistency is king, let's keep it to be the same names as the Rust versions

src/ffi/wallet.rs
```diff
@@ -0,0 +1,150 @@
...
+) {
+    catch_unwind_cb(user_data, o_cb,  || -> ResultReturn<()> {
+        let user_data = OpaqueCtx(user_data);
+        let wallet_xorurl = (*app).wallet_create()?;
+        let key_url_str = from_c_str_to_string_option(key_url);
+        let secret_key_str = from_c_str_to_string_option(secret_key);
+        let name_str = from_c_str_to_string_option(name);
+        let preload_str = from_c_str_to_string_option(preload);
+        let pay_with_str = from_c_str_to_string_option(pay_with);
+        if !no_balance {
```
bochaco: You shouldn't do any of this, you need to simply call the equivalent one from Rust API, this applies to absolutely all the bindings you are creating , thee is 1-to-1 mapping, only 1 FFI function per each Rust function, no more, no less. All validations are done by the Rust API, you only convert from C repr to Rust native data types and pass them down, no additional logic here at al. This is very important, since any change in logic should happen only in Rust API impl, no binding will need to be changed for any logic change, FFI/bindings will only change if the signature of an API changes.

src/ffi/wallet.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        let wallet_xor_url = to_c_str(wallet_xorurl)?;
+        o_cb(user_data.0, FFI_RESULT_OK, wallet_xor_url.as_ptr());
+        Ok(())
+    })
+}
+
+
+#[no_mangle]
+pub unsafe extern "C" fn insert_balance_to_wallet(
```
bochaco: Consistency is king, let's keep it to be the same names as the Rust versions

src/ffi/wallet.rs
```diff
@@ -0,0 +1,150 @@
...
+        };
+        let the_name = (*app).wallet_insert(&target_str, name_str, set_default, &sk)?;
+        let result_name = to_c_str(the_name)?;
+        o_cb(user_data.0, FFI_RESULT_OK, result_name.as_ptr());
+        Ok(())
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn query_wallet_balance(
```
bochaco: Consistency is king, let's keep it to be the same names as the Rust versions

src/ffi/wallet.rs
```diff
@@ -0,0 +1,150 @@
...
+        let target_str = from_c_str(target)?;
+        let balance = (*app).wallet_balance(&target_str)?;
+        let amount_result = to_c_str(balance)?;
+        o_cb(user_data.0, FFI_RESULT_OK, amount_result.as_ptr());
+        Ok(())
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn transfer_wallet_balance(
```
bochaco: Consistency is king, let's keep it to be the same names as the Rust versions

src/ffi/xorurl.rs
```diff
@@ -0,0 +1,87 @@
...
+use super::helpers::{from_c_str_to_str_option, to_c_str};
+use ffi_utils::{catch_unwind_cb, from_c_str, FfiResult, OpaqueCtx, FFI_RESULT_OK};
+use safe_core::ffi::arrays::{XorNameArray};
+use safe_nd::{XorName};
+use std::os::raw::{c_char, c_void};
+use super::ffi_structs::XorUrlEncoder;
+
+// todo: Can be convertered to a struct
+#[no_mangle]
+pub unsafe extern "C" fn encode_xor_url(
```
bochaco: Consistency is king, look at my previous comment further above, how about `xorurl_encode`

src/ffi/xorurl.rs
```diff
@@ -0,0 +1,87 @@
...
+        let encoded_xor_url = NativeXorUrlEncoder::encode(xor_name, type_tag, data_type_enum, content_type_enum, url_path, Some(vec![]), Some(content_version), &encoding_base)?; //todo: update sub_names parameter
+        let encoded_string = to_c_str(encoded_xor_url)?;
+        o_cb(user_data.0, FFI_RESULT_OK, encoded_string.as_ptr());
+        Ok(())
+    })
+}
+
+// todo: Can be convertered to a struct
+#[no_mangle]
+pub unsafe extern "C" fn new_xor_url_encoder(
```
bochaco: Same comment as above, let's keep some consistency, it's very important

src/ffi/xorurl.rs
```diff
@@ -0,0 +1,87 @@
...
+        let url_path = from_c_str_to_str_option(path);
+        let encoder = NativeXorUrlEncoder::new(xor_name, type_tag, data_type_enum, content_type_enum, url_path, Some(vec![]), Some(content_version))?; //todo: update sub_names parameter
+        let ffi_encoder = encoder.into_repr_c()?;
+        o_cb(user_data.0, FFI_RESULT_OK, &ffi_encoder);
+        Ok(())
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn new_xor_url_encoder_from_xor_url(
```
bochaco: Here you don't even need the `new_` prefix I think, `xorurl_encoder_from_url` is already good enough and with some consistency with Rust versions

src/lib.rs
```diff
@@ -9,6 +9,16 @@
 #[macro_use]
 extern crate validator_derive;
 
-mod api;
+extern crate ffi_utils;
```
bochaco: I think this file shouldn't be the entry ponit for FFI lib and therefore this file shouldn't need any change

src/main.rs
```diff
@@ -9,6 +9,7 @@
 mod api;
 mod cli;
 mod subcommands;
+mod ffi;
```
bochaco: We shouldn't need this here

src/subcommands/auth.rs
```diff
@@ -8,7 +8,7 @@
 
 use dirs;
 use log::debug;
-use safe_cli::Safe;
+use crate::api::Safe;
```
bochaco: We should depend on the crate not on a module, the cli is just a Rust app, i should be able to even get the CLI app from this repo and move it to another repo without needing to change the code, only its Cargo.toml will need to have safe-api dependency and that's all.

src/subcommands/cat.rs
```diff
@@ -11,7 +11,7 @@ use super::OutputFmt;
 use crate::subcommands::auth::auth_connect;
 use log::debug;
 use prettytable::Table;
-use safe_cli::{NrsMapContainerInfo, Safe, SafeData};
+use crate::api::{NrsMapContainerInfo, Safe, SafeData};
```
bochaco: Same as above, CLI is an app which depends on safe-api crate, not on a mod

src/subcommands/fake_auth.rs
```diff
@@ -7,7 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use log::debug;
-use safe_cli::Safe;
+use crate::api::Safe;
```
bochaco: Same as above, CLI is an app which depends on safe-api crate, not on a mod

src/subcommands/files.rs
```diff
@@ -9,7 +9,7 @@
 use super::helpers::get_from_arg_or_stdin;
 use super::OutputFmt;
 use prettytable::{format::FormatBuilder, Table};
-use safe_cli::{Safe, XorUrl, XorUrlEncoder};
+use crate::api::{Safe, XorUrl, XorUrlEncoder};
```
bochaco: Same as above, CLI is an app which depends on safe-api crate, not on a mod

src/subcommands/helpers.rs
```diff
@@ -7,7 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use log::debug;
-use safe_cli::XorName;
+use crate::api::XorName;
```
bochaco: Same as above, CLI is an app which depends on safe-api crate, not on a mod

src/subcommands/keys.rs
```diff
@@ -10,7 +10,7 @@ use super::auth::auth_connect;
 use super::helpers::{get_from_arg_or_stdin, get_secret_key};
 use super::OutputFmt;
 use log::{debug, warn};
-use safe_cli::{BlsKeyPair, Safe};
+use crate::api::{BlsKeyPair, Safe};
```
bochaco: Same as above, CLI is an app which depends on safe-api crate, not on a mod

src/subcommands/nrs.rs
```diff
@@ -9,7 +9,7 @@
 use super::helpers::get_from_arg_or_stdin;
 use super::OutputFmt;
 use prettytable::{format::FormatBuilder, Table};
-use safe_cli::{Safe, XorUrl};
+use crate::api::{Safe, XorUrl};
```
bochaco: Same as above, CLI is an app which depends on safe-api crate, not on a mod

src/subcommands/wallet.rs
```diff
@@ -12,7 +12,7 @@ use super::helpers::{get_from_arg_or_stdin, get_secret_key};
 use super::keys::{create_new_key, print_new_key_output};
 use super::OutputFmt;
 use log::debug;
-use safe_cli::{BlsKeyPair, Safe};
+use crate::api::{BlsKeyPair, Safe};
```
bochaco: Same as above, CLI is an app which depends on safe-api crate, not on a mod

src/api/mod.rs
```diff
@@ -8,21 +8,21 @@
 
 mod auth;
 mod constants;
-mod errors;
+pub mod errors;
```
bochaco: We shoudln't need to expose all this mods as public, we expose only some parts of it and that's done from the crate's main entry point, the lib.rs, we should keep that as it is. Whatever is needed by FFI lib should be exposed from an FFI specific module/files

src/ffi/keys.rs
```diff
@@ -0,0 +1,94 @@
...
+        xor_url: *const c_char,
+        safe_key: *const BlsKeyPair,
+        pre_load: *const c_char)
+) {
+    catch_unwind_cb(user_data, o_cb,  || -> ResultReturn<()> {
+        let user_data = OpaqueCtx(user_data);
+        let preload_str = from_c_str_to_string_option(preload);
+        let pay_with_str = from_c_str_to_string_option(pay_with);
+        let pk_with_str = from_c_str_to_string_option(pk);
+        let (xorurl, key_pair, _amount) = if test_coins {
```
bochaco: None of this should be done here, no validations or any logic, the only logic you need here is to convert from C repr into/from Rust native types. All these validations should be removed

src/ffi/keys.rs
```diff
@@ -0,0 +1,94 @@
...
+        user_data: *mut c_void,
+        result: *const FfiResult,
+        balance: *const c_char)
+) {
+    catch_unwind_cb(user_data, o_cb,  || -> ResultReturn<()> {
+        let user_data = OpaqueCtx(user_data);
+        let key_url = from_c_str(key)?;
+        let secret_key = from_c_str_to_string_option(secret);
+        let sk = secret_key.unwrap_or_else(|| String::from(""));
+        let current_balance = if key_url.is_empty() {
```
bochaco: None of this should be done here, no validations or any logic, the only logic you need here is to convert from C repr into/from Rust native types. Any other logic should be removed

src/ffi/wallet.rs
```diff
@@ -0,0 +1,150 @@
...
+        name: *const c_char)
+) {
+    catch_unwind_cb(user_data, o_cb,  || -> ResultReturn<()> {
+        let user_data = OpaqueCtx(user_data);
+        let target_str = from_c_str(target)?;
+        let secret_key_str = from_c_str_to_string_option(secret_key);
+        let name_str = from_c_str_to_string_option(name);
+        let key_url_str = from_c_str_to_string_option(key_url);
+
+        let sk = match key_url_str {
```
bochaco: None of this should be done here, no validations or any logic, the only logic you need here is to convert from C repr into/from Rust native types. Any other logic/validation should be removed

build.rs
```diff
@@ -0,0 +1,181 @@
+// Copyright 2019 MaidSafe.net limited.
```
bochaco: Can we move this file onto ffi folder?

build.rs
ravinderjangra: Currently, bindgen reads only from `lib.rs` file so that's why we are using this one. Once we have some structure change sin repo then maybe we can simplify this

src/api/xorurl.rs
ravinderjangra: I was thinking the same but the `XorUrlEncoder` struct's fields are private and that's why I can't implement this in ffi file. If there is no problems if we make those fields pub then this can be sorted I think.

src/main.rs
ravinderjangra: The code wouldn't build without adding this here: https://www.reddit.com/r/rust/comments/4wi1nc/am_i_misunderstanding_rusts_module_system/d676042/

build.rs
ravinderjangra: I don't we can move this for now because it's used when building the package. but if we separate all three `cli`, `api`, and `ffi` as separate creates/libs/bins then maybe we can move this to the ffi one.

src/api/xorurl.rs
```diff
@@ -63,6 +93,24 @@ impl std::fmt::Display for SafeDataType {
...
+            0 => SafeDataType::SafeKey,
+            1 => SafeDataType::PublishedImmutableData,
+            2 => SafeDataType::UnpublishedImmutableData,
+            3 => SafeDataType::SeqMutableData,
+            4 => SafeDataType::UnseqMutableData,
+            5 => SafeDataType::PublishedSeqAppendOnlyData,
+            6 => SafeDataType::PublishedUnseqAppendOnlyData,
+            7 => SafeDataType::UnpublishedSeqAppendOnlyData,
+            8 => SafeDataType::UnpublishedUnseqAppendOnlyData,
+            _ => panic!("Unknown value: {}", value),
```
bochaco: `panic!` is forbidden ;)

bochaco:
This is a fantastic job here @ravinderjangra ! really good stuff, it seems this PR puts us in a very nice position to start exposing the API to other languages.

PR #256, opened 2019-09-26
fix: cat without auth

b-zee:
Dependency was fixed (https://github.com/maidsafe/safe_client_libs/issues/899) to allow `cat` without auth.

### DIFF
--- a/safe_cli/subcommands/cat.rs
+++ b/safe_cli/subcommands/cat.rs
@@ -8,7 +8,7 @@
 
 use super::helpers::{get_from_arg_or_stdin, xorname_to_hex};
 use super::OutputFmt;
-use crate::subcommands::auth::auth_connect;
+use crate::subcommands::auth::connect_without_auth;
 use log::debug;
 use prettytable::Table;
 use safe_api::{NrsMapContainerInfo, Safe, SafeData};
@@ -32,9 +32,7 @@ pub fn cat_commander(
     let url = get_from_arg_or_stdin(cmd.location, None)?;
     debug!("Running cat for: {:?}", &url);
 
-    // TODO: pending: https://github.com/maidsafe/safe_client_libs/issues/899
-    // switch to connect_without_authL: connect_without_auth(safe)?;
-    auth_connect(safe)?;
+    connect_without_auth(safe)?;
     let content = safe.fetch(&url)?;
     match &content {
         SafeData::FilesContainer {


b-zee:
I made a mistake during the pull request. I forgot to make an edit on the GitHub site. I fixed it however, and force-pushed the amended commit, but the test doesn't seem to run. Anything I could do here to make it verify the test?

When I pushed the commit, all the tests ran successfully though.

bochaco:
When you force-push ot should run CI again (ignore Travis here as we are only maintaining Jenkins CI now). Can you please try resolving that conflict (we just merged a PR and is causing that conflict now) and let's see if the CI runs ok?

b-zee:
Thanks @bochaco, I see I made another mistake indeed, I didn't pull the last master. Anyhow, I fixed that too, but the last build check failed too. Locally the test works fine though. I'm not sure how to fix this.

bochaco:
nw @b-zee ! I presume you are locally running only the tests using the scl-mock, which is just a fake we have in the crate, but in CI we run that plus run them against mock-network and using a real safe_auth CLI to authorise the cli app. So, when using the real authorisation mechanism against mock-network the permissions are effectively enforced by SCL lib, so here you see it's failing due to access denied error: https://jenkins.maidsafe.net/blue/organizations/jenkins/pipeline-safe_cli/detail/PR-256/4/pipeline#step-67-log-660. You wouldn't get that error with the fake mock since we don't check for permisions or anything like that.

Let me check if from my understanding this is effectively possible to do now, i.e. `cat`-ing without logging in

b-zee:
Don't worry about it. I'll re-check this tomorrow, then it might be the case I haven't tested this properly with my local vault yet...

bochaco:
Ah yeah, if you test with local vault you should reproduce the same issue as in CI

bochaco:
Ok, some more info :) the test failing I think it's expected to fail since it's trying to cat a Wallet. Wallets are UnpublishedData, therefore only accessible to the owner, so you cannot `cat` a wallet without authorising yourself with the Wallet owner's account. 
It seems we'll need a bit of more logic to proerly do this:
1- check if we have credentials stored already to connect with, if so then connect and fetch data as usual/now
2- if there is no credentials, then connect without auth and:
a) either try to fetch the data and handle the error if the data being fetch is Unpublished showing an error saying that it needs to be authorised first before being able to fetch that data type
b) use the XorUrlEncoder::from_url to check upfront what kind of data type it is, and either fetch it (if it's PublishedData) or error upfront without trying to fetch (if it's UnpublishedData).

I think even that 2.a is more costly in terms of going to the network is the one I'd prefer since we don't have to maintain that logic of 2.b checking for different types of data on the CLI app.

b-zee:
Thanks @bochaco! Your proposed solution makes sense.

I think the `AccessDenied` error should tell the user just that: 'access denied, perhaps try to authenticate?'  This could be a generic error handler so all cases of this error in the CLI will be displayed to the user in a meaningful way.

The solution you propose to authenticate the user depending on the availability of credentials is a good one. I can imagine in the future there will be more ways to provide the credentials (environment variable, `--token` argument) so perhaps this can also be generalized. If the credentials are provided, then _always_ authenticate the CLI with those credentials. This is not specific to `cat`.

bochaco:
I really like all that @b-zee , yes, it's good if we work on generalising it both aspects as you say, yes! that should cover some use cases in the future with other type of restricted access to non public content.

bochaco:
@b-zee just checking if you are still interested in this implementation, otherwise I may give it a try in the next few days, I think it's a good moment to allow CLI to work in read-only mode in every situation possible. Please let me know, I don't wanna discourage you just by starting with this if you are still working on it or intending to.

b-zee:
Yes, sorry, I do not intend to work on this. I'm not comfortable enough with the code and how the procedures are supposed to work. Thanks a lot for your effort in advance!

bochaco:
Ok, no worries, please don't hesitate to jump in to review my PR then if you feel like ofc :)

bochaco:
Closing this as agreed, and it's being implemented by #356

PR #260, opened 2019-09-30
Expose XorNameArray and ProcessedEntries structs bindings

ravinderjangra:


safe-ffi/ffi/ffi_structs.rs
```diff
@@ -12,11 +13,17 @@ use safe_api::xorurl::{SafeContentType, SafeDataType, XorUrlEncoder as NativeXor
 use safe_api::{
     BlsKeyPair as NativeBlsKeyPair, NrsMapContainerInfo as NativeNrsMapContainerInfo, ResultReturn,
 };
-use safe_core::ffi::arrays::XorNameArray;
 use safe_nd::XorName;
 use std::ffi::CString;
 use std::os::raw::c_char;
 
+/// Constant byte length of `XorName`.
+pub const XOR_NAME_LEN: usize = 32;
```
m-cat: Can we take this from `safe-nd` instead of redefining it here?

safe-ffi/build.rs
```diff
@@ -24,6 +24,8 @@ mod bindings {
     use std::path::Path;
     use unwrap::unwrap;
 
+    const BLS_PUBLIC_KEY_LEN: usize = 48;
```
m-cat: Possible to import this from `threshold_crypto` instead? i.e. see [safe-core](https://github.com/maidsafe/safe_client_libs/blob/b96478fbe1c49f2f0876dcea8e189405426fb503/safe_core/src/ffi/arrays.rs#L18).

ravinderjangra: But we don't have `threshold_crypto` as a dependency so I don't think adding this just for a constant is required. But if it doesn't matter then I am more than happy to update this.

m-cat: Ah, thought that it was a dependency. Agreed. Wrote [this issue](https://github.com/maidsafe/safe-nd/issues/118) in the meantime.

safe-ffi/ffi/ffi_structs.rs
```diff
@@ -3,6 +3,7 @@ use ffi_utils::{from_c_str, vec_into_raw_parts};
 use safe_api::files::{
     FileItem as NativeFileItem, FilesMap as NativeFilesMap, ProcessedFiles as NativeProcessedFiles,
 };
+use safe_api::nrs::ProcessedEntries as NativeProcessedEntries;
```
bochaco: You should expose ProcessedEntries from `safe-api/src/api/mod.rs` and then you depend on that one

ravinderjangra: Do you mean something like this: https://github.com/maidsafe/safe-cli/blob/87a3d737fb60be30503de45628b691e35843f0f3/safe-api/src/api/mod.rs#L31

ravinderjangra: Just to clarify I'll still have to make `ProcessedEntries` public?

bochaco: yes correct, otherwise you are pulling it from some implementation detail rather than from the exposed API/types, right now you can say it's a coincidence they are called the same (what we expose from mod.rs and what's is internally called in nrs.rs), so ffi always depends on what is exposed from the mod.rs and nothing more, that's the API and all it should care about

safe-ffi/Cargo.toml
```diff
@@ -18,6 +18,7 @@ build = "build.rs"
 edition = "2018"
 
 [lib]
+name = "safe_api"
```
bochaco: @ravinderjangra I think we should keep the name safe_ffi as this causes a collition as thrown by cargo/rust when building from root path of repo:
```
$ cargo build --features bindings
warning: output filename collision.
The lib target `safe_api` in package `safe-ffi v0.4.0 (/home/bochaco/safe/bochaco-safe-cli/safe-ffi)` has the same output filename as the lib target `safe-api` in package `safe-api v0.4.0 (/home/bochaco/safe/bochaco-safe-cli/safe-api)`.
Colliding filename is: /home/bochaco/safe/bochaco-safe-cli/target/debug/libsafe_api.rlib
The output filenames should be unique.
```

ravinderjangra: I'll change that. I was thinking to keep it simple for the native libs, no worries we can rename these files in C# side. I was building from the `safe-ffi` and never noticed this.

bochaco: I don't think you shouldn't do that either @ravinderjangra , as in the future we could be generating a .so (`dylib` type) for Rust safe-api lib as well

PR #261, opened 2019-10-01
Update fetch and refactor to use json for nrs_map and files_map

ravinderjangra:


safe-ffi/ffi/ffi_structs.rs
```diff
@@ -354,6 +367,20 @@ pub struct NrsMapContainerInfo {
     pub data_type: u64,
 }
 
+impl NrsMapContainerInfo {
+    pub fn new_nrs_map_containet_info() -> ResultReturn<NrsMapContainerInfo> {
```
m-cat: Why not just call this function `new`, as it will be called like `NrsMapContainerInfo::new()`? Also, you can have it return `ResultReturn<Self>`.

safe-ffi/ffi/ffi_structs.rs
```diff
@@ -354,6 +367,20 @@ pub struct NrsMapContainerInfo {
...
+        Ok(NrsMapContainerInfo {
```
m-cat: This line can be `Ok(Self {`.

safe-ffi/ffi/ffi_structs.rs
```diff
@@ -354,17 +489,49 @@ pub struct NrsMapContainerInfo {
...
+            }
+
+            if !self.nrs_map.is_null() {
+                let _ = CString::from_raw(self.nrs_map as *mut _);
+            }
+        }
+    }
+}
+
+impl NrsMapContainerInfo {
```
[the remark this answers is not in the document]
ravinderjangra: Actually that's how I started first but then the team wanted to keep all FFI stuff seperate, then I moved the code in FFI and all the structs used by bindings are defined in ffi_structs file.

PR #263, opened 2019-10-01
"helpers"

m-cat:
+ Remove `to_c_str`:

I was surprised to see this function because it's not something we use in FFI
elsewhere (like SCL). It's basically a wrapper around `CString::new` so as far
as I can tell it obscures the actual function call for no real benefit.

+ Remove panics from helpers

+ Remove `from_c_str_to_string_option`:

Where this is needed (which right now it's not), this can be replaced with `from_c_str(ptr).ok()`. As for `#[inline]`, I'm not sure if it's necessary: https://doc.rust-lang.org/reference/attributes/codegen.html#the-inline-attribute


ravinderjangra:
It is better. I am not sure about removing `from_c_str_to_string_option` Because it maybe used for some APIs but it seems no API is currently using this so in case if we need this in future we can add this later.

PR #267, opened 2019-10-05
Showcase: Organisation of data API code

oetyng:
## This draft PR serves as a showcase.

_The code in this PR is to be considered pseudo-code, as no effort has been put into ensuring it compiles. There are likely plenty of idiomatic errors and similar, the focus being on the structure, systematization and naming._

The intention of this PR is to demonstrate opportunities of simplification, improved structure and readability of the rust code base.

The chosen substrate for the showcase is the `safe_client_libs.rs` file. It serves as a good example, as it is a file with many lines of code, a lot of functionality squeezed into one format, and as a result a lot of bloat in member naming.

-----------

## Naming

First of all, some concepts have been straightened out.
There is an inflation of the use of words such as "safe" and "app", and the combination of them is hardly making any sense. The code is within a safe repository (preferably named `safe-api`) and that is guidance enough for developers that we are dealing with `safe`. When drilling down into the layers, we assume everyone is aware of this, and omit any further references to "safe".

Doing this, now opens up opportunities for being more descriptive and accurate in our naming in subsequent layers. As an example, `SafeApp` is renamed to `Session`, which actually has an inherent meaning and informs developers of the intended usage. A lot more information has been conveyed, and confusion avoided.

Correspondingly, `SafeAppError` is renamed to `SessionError`. A number of similar adjustments have been made,

## Separation of concerns

All operations pertinent to a specific data type, has been placed in its own file.

Instead of placing everything within one struct/impl-combo in one file, we identify groups of functionality, and place them in separate struct/impl-combos within their respective files.
This way, we do not need to prefix method names with distinguishing information.

`safeApp.files_get_published_immutable` becomes `immutable_data.get`

More specifically, you'd do something like:

```
var session = SessionFactory.InitSession();
var networkApis = new NetworkData(session);
var imDataApi = networkApis.PubImmutableData;

var data = imDataApi.get(xorName);
```

## Layering

As an organizing structure, we use `NetworkData` struct, to be our reference to the three different data APIs; 

- `PubSeqAppendOnlyData`
- `PubImmutableData`
- `MutableData`

Each of these are implemented and tested in their respective files.

A `NetworkData` instance is initialized passing a `Session` instance to its ctor. The initialization of the API instances is encapsulated within it.

## Consistency
As an example, for different ways of referring to the same thing, in the error messages these were used:

- `mutable data`
- `MD`
- `SeqMD`
- `Sequenced MutableData`

All were replaced with `Sequenced MutableData`, as it is the full description, no shorts, no part of the name left out. We don't need to save space in error messages.

Consistency is very important for new people assimilating a code base. Slightest deviation, irregularity or uncertainty is a potential pattern - it might have a meaning and a significance that the user has to track, identify and categorize. When everything is new, inconsistencies make it order of magnitudes harder to spot and grok the _actual_ patterns, concepts and important things.

It is important at every level, from the simplest error messages, up to what structure a system has, the strategies used to solve problems. These things multiply and cross-pollinate across levels very easily.
 
Not only for new people, but for extending and maintaining code, consistency leaves much less room for mistakes, misinterpretation and lowers the cognitive load associated with introducing changes. There is one more important reason, I would say, to _love consistency_ and treat it like a _very_ important part of a system: without it creativity is thwarted, insights about what the system does and _can_ do is hampered and the full creative potential of developers is neither nourished nor taken well care of. More time is spent avoiding to touch code, trying to grasp the essentials of the system and struggling with fully groking the consequences of change.

There will always be compromises in any systematization. This is not about idealistic puritanism, just about being as consious as absolutely possible in every choice and step, which also makes the unevitable compromises so much clearer and easier to identify and handle.

## Compromises:

We are here calling the constructors of other classes from within a constructor. This is generally not considered a good practice. Instead we'd prefer injecting the instances.

```
public NetworkData(IPubSeqAppendOnlyData pubSeqAppendOnlyData, IPubImmutableData pubImmutableData, IMutableData mutableData)
{
...
}
```
 
The types are here shown with typical C# naming, prefixing interfaces with a capital i,
In rust, we would be using traits. The naming conventions for traits when utilizing them in the capacity of interfaces has not been investigated prior to this showcase.

Dependency injection allows us to exert inversion of control, which is a formidable tool in building code which is **testable**, **extensible** and **reusable**. In my opinion, the most important of those characteristics being **testable**, and the least important **reusable**, since that is often much less feasible in practice than often portrayed in theory.

--------------------------

This draft will be continuously updated, both in description and code. Consider it a work in progress.


joshuef:
In general this all sounds grand! 👍 

A couple of comments, re: things going on at the moment in the API:

a) wrapping SCL. We do this to provide easier APIs for use elsewhere (ie, this started to abstract away SCL for CLI). Here: `safeApp.files_get_published_immutable becomes immutable_data.get` sounds great.

But b) We're also trying to make _clearer_ APIs too. So while I think the suggestion `safeApp.files_get_published_immutable becomes immutable_data.get` is grand. It shouldn't be at the cost of simplifying things so devs can come and not even have to worry about the concepts of immutable data... `files.get` being another API eg. effectively wrapping our Immutable data APIs.


also:

```
    mutable data
    MD
    SeqMD
    Sequenced MutableData
```

I think we can have Unsequenced MutableData also (i think... i lose track sometimes though).

oetyng:
A new take on the suggestion included in this PR needs to be done after CRDT data types are in place.

PR #270, opened 2019-10-07
feat(cli): new dog command to inspect content retrieving only metadata

### LINKED ISSUE #244: feature request - return file info without attempting to render file
Currently information about a file can be obtained by `safe cat <xor_url> --info`. However the `cat` operation attempts to render the file in question. This means for non text files there is frequently a large output to the terminal and the file metadata (which is displayed first) cannot be viewed.

Eg:
image

It would be useful to have some option to _only_ return the file information to get an output like:
```
> ./safe cat safe://hbkyxynsiqqka6bz7g333sufp1euu9kiaumobcpot5kmijbt9kfbxyfwqw --info
Native data type: ImmutableData (published)
XOR name: 0x5aae72b1e0dfa6ce736995b244e7f557135c02c6c23b52ea90c7ea285e02d1d4
Media type: application/vnd.ms-excel
```

maybe using a different flag than `--info`?

bochaco:
Resolves #244

### DIFF
--- a/safe-cli/tests/cli_dog.rs
+++ b/safe-cli/tests/cli_dog.rs
@@ -1,0 +1,101 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+mod common;
+
+#[macro_use]
+extern crate duct;
+
+use common::{
+    create_preload_and_get_keys, get_bin_location, get_random_nrs_string,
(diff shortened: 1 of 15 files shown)


safe-cli/tests/cli_dog.rs
```diff
@@ -0,0 +1,101 @@
...
+        "nrs",
+        "create",
+        &nrsurl,
+        "-l",
+        &container_xorurl,
+    )
+    .read()
+    .unwrap();
+
+    let cat_output = cmd!(get_bin_location(), "dog", &nrsurl, "--json",)
```
joshuef: there's a couple instances of still referring to `cat` here.

safe-cli/tests/cli_dog.rs
```diff
@@ -0,0 +1,101 @@
...
+        "nrs",
+        "create",
+        &nrsurl,
+        "-l",
+        &safekey_xorurl,
+    )
+    .read()
+    .unwrap();
+
+    let cat_output = cmd!(get_bin_location(), "dog", &nrsurl, "--json",)
```
joshuef: here too

bochaco:
@joshuef I've seen some network activity when it builds `rust_sodium-sys`, and even failed for me when my network connection was bad, could that be your case now?

PR #275, opened 2019-10-10
refactor(ffi): impl Drop for xorurls and use into_raw()

ravinderjangra:


safe-ffi/ffi/fetch.rs
```diff
@@ -32,8 +32,9 @@ pub unsafe extern "C" fn fetch(
                 media_type,
             } => {
                 let (data, data_len, data_cap) = vec_into_raw_parts(data.to_vec());
+                let xorurl_c_str = CString::new(xorurl.clone())?;
                 let published_data = PublishedImmutableData {
-                    xorurl: CString::new(xorurl.clone())?.as_ptr(),
+                    xorurl: xorurl_c_str.into_raw(),
```
m-cat: Just FYI: We don't need the temporary variables here, like we did when passing pointers to `o_cb`. In that case, we needed the pointed-to data to live long enough, but in this case we know the data won't get dropped because `into_raw` *consumes* `xorurl_c_str`. So it's up to you whether you feel the temporary `xorurl_c_str` var is necessary here; in SCL we avoid the extra verbosity and would do `CString::new(xorurl.clone())?.into_raw()`.

ravinderjangra: Will update this.

PR #279, opened 2019-10-15
fix(ffi): fix keys_create to return new BlsKeyPair instance when None…

ravinderjangra:
… returned from API

### DIFF
--- a/safe-ffi/ffi/keys.rs
+++ b/safe-ffi/ffi/keys.rs
@@ -44,12 +44,24 @@ pub unsafe extern "C" fn keys_create(
         let pk_option = from_c_str_to_str_option(pk);
         let (xorurl, keypair) = (*app).keys_create(from_option, preload_option, pk_option)?;
         let xorurl_c_str = CString::new(xorurl.to_string())?;
-        o_cb(
-            user_data.0,
-            FFI_RESULT_OK,
-            xorurl_c_str.as_ptr(),
-            &bls_key_pair_into_repr_c(&keypair.as_ref().unwrap())?,
-        );
+        match keypair {
+            Some(keypair) => o_cb(
+                user_data.0,
+                FFI_RESULT_OK,
+                xorurl_c_str.as_ptr(),
+                &bls_key_pair_into_repr_c(&keypair)?,
+            ),
+            None => o_cb(
+                user_data.0,
+                FFI_RESULT_OK,
+                xorurl_c_str.as_ptr(),
+                &BlsKeyPair {
+                    pk: CString::new(from_c_str(pk)?)?.into_raw(),
+                    // null because we only have the pk, todo: maybe expose another callback to manage this in future
+                    sk: std::ptr::null(),
+                },
+            ),
+        };
         Ok(())
     })
 }
--- a/safe-ffi/ffi/keys.rs
+++ b/safe-ffi/ffi/keys.rs
@@ -44,23 +44,19 @@ pub unsafe extern "C" fn keys_create(
         let pk_option = from_c_str_to_str_option(pk);
         let (xorurl, keypair) = (*app).keys_create(from_option, preload_option, pk_option)?;
         let xorurl_c_str = CString::new(xorurl.to_string())?;
-        match keypair {
-            Some(keypair) => o_cb(
-                user_data.0,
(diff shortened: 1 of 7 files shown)


safe-ffi/ffi/keys.rs
```diff
@@ -44,12 +44,24 @@ pub unsafe extern "C" fn keys_create(
...
+                &bls_key_pair_into_repr_c(&keypair)?,
+            ),
+            None => o_cb(
+                user_data.0,
+                FFI_RESULT_OK,
+                xorurl_c_str.as_ptr(),
+                &BlsKeyPair {
+                    pk: CString::new(from_c_str(pk)?)?.into_raw(),
+                    // null because we only have the pk, todo: maybe expose another callback to manage this in future
+                    sk: std::ptr::null(),
```
m-cat: Hmm, we shouldn't be using null pointers as the FFI equivalent of `String`. A null pointer represents the `None` variant of an `Option` - see [here](https://github.com/maidsafe/safe_client_libs/wiki/FFI-quirks-and-utils#summary). So either change the native `BlsKeyPair` to carry `Option<String>`, or pass an empty string here (which still needs to be converted the normal way through `CString::new`).

ravinderjangra: I'll change this to use an empty string and then convert that in `CString`

m-cat: > I'm not using the native BlsKeyPair here, it's an FFI struct.

What I meant is, the fields in an FFI struct correspond to, and are C equivalents of, the fields of the native struct. So if we're allowing null pointers for an FFI field, the corresponding type in the native struct should be an `Option<_>`.

ravinderjangra: @m-cat, is this implementation correct? :
```
&BlsKeyPair {
    pk: CString::new(from_c_str(pk)?)?.into_raw(),
    sk: CString::new(String::new())?.into_raw(),
},
```

ravinderjangra: I'll update this later to use different callbacks based on the use case. This the only place where the functionality is different.

bochaco: I've see in fetch you are using diff callbacks, and here you say you'd do the same, why are you doing such a thing? in fetch perhaps i get it that it's more complex structure to create the ffi, but here i don't see why you'd need two diff callbacks, @ravinderjangra can you please explain what was the reasoning behind this?

safe-ffi/ffi/keys.rs
```diff
@@ -44,12 +44,23 @@ pub unsafe extern "C" fn keys_create(
...
+            None => o_cb(
+                user_data.0,
+                FFI_RESULT_OK,
+                xorurl_c_str.as_ptr(),
+                &bls_key_pair_into_repr_c(&NativeBlsKeyPair {
+                    pk: from_c_str(pk)?,
+                    sk: String::new(),
+                })?,
+            ),
+        };
```
m-cat: Hope you don't mind that I rewrote your code a bit, but this can be rewritten to be a bit shorter and simpler:

```rust

        let keypair = match keypair {
            Some(keypair) => keypair,
            None => NativeBlsKeyPair {
                pk: from_c_str(pk)?,
                sk: String::new(),
            }?,
        };
        o_cb(
            user_data.0,
            FFI_RESULT_OK,
            xorurl_c_str.as_ptr(),
            &bls_key_pair_into_repr_c(&keypair)?,
        );
```

safe-ffi/ffi/keys.rs
ravinderjangra: because for this API, in one case we are getting the BlsKeyPair and in another, we'are not. So, either we can use the method (same as this PR) or I can use different callbacks. I'm fine with this one as well.

safe-ffi/ffi/keys.rs
ravinderjangra: I didn't think of that. It's much easy to read and simple as well. Will update to use this now.

bochaco: I like that, I usually tend to do this so it's very clear what is the diff in each case, e.g. here the diff is only the keypair is different in each scenario and not the rest of args passed to the callback

PR #290, opened 2019-10-21
refactor(ffi): create an FfiError struct and move out from api all Error traits impl which are required for FFI only

bochaco:


bochaco:
@m-cat I was also thinking if we need the code in sapi, I think they can stay there if we use them as constants, like in this PR, so whenever we add a new code it's made from api bottom up to ffi. The wrapper is needed only to be able to implement some traits which are used by ffi-utils, if we get rid of those needs we shouldn't even need FfiError wrapper, but ffi-utils is being used all over in FFI code right now

bochaco:
I'm merging this as it's a step forward in any case, so please send a PR @m-cat if you think you can remove move error codes from api to ffi. This PR gets rid of the last bit of ffi code which was mixed in api.

m-cat:
Of course @bochaco , no need to delay this PR over that concern. I wrote an issue, I'll play around with the errors later because it may have some relevance to how we handle them in SCL.

PR #296, opened 2019-10-23
refactor/ffi: move error codes to FFI Error

### LINKED ISSUE #291: Is FfiError necessary?
I'm just wondering what the purpose of FfiError is, as it's currently a thin wrapper around Error and doesn't seem necessary to me.

I think it would make more sense if the error codes were moved from Error to FfiError, if possible. 

In SCL, we don't have separate errors for FFI, i.e. we have just `AppError` for everything in `safe-app`. If you're able to get a better separation for these errors here though, SCL can probably adopt a similar strategy.

m-cat:
This PR provides a nicer separation of concerns than previously. Now FFI-specific error codes do not appear outside the `ffi` module. The organization of the code is also simplified with respect to the Native and FFI error types.

Resolves #291

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2343,6 +2343,7 @@ name = "safe-ffi"
 version = "0.5.0"
 dependencies = [
  "ffi_utils 0.13.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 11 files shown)


bochaco:
There is a problem on CLI tests on Mac that will need to be taken care of @m-cat

PR #306, opened 2019-11-06
chore(ffi): use new XorUrlEncoder instead of null ptr in parse_and_resolve_url

ravinderjangra:

### DIFF
--- a/safe-ffi/ffi/ffi_structs.rs
+++ b/safe-ffi/ffi/ffi_structs.rs
@@ -165,6 +165,21 @@ impl Drop for XorUrlEncoder {
     }
 }
 
+impl XorUrlEncoder {
+    pub fn new() -> Result<Self> {
+        Ok(Self {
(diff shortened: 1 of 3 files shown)


safe-ffi/ffi/ffi_structs.rs
```diff
@@ -165,6 +165,21 @@ impl Drop for XorUrlEncoder {
...
 
+impl XorUrlEncoder {
+    pub fn new() -> Result<Self> {
+        Ok(Self {
+            encoding_version: 0,
+            xorname: [0; 32],
+            type_tag: 0,
+            data_type: 0,
+            content_type: 0,
+            path: CString::new(String::new())?.into_raw(),
```
m-cat: It should be safe to `unwrap` here, `CString::new` won't ever fail for an empty string. This way you can avoid returning a `Result` here and just return `Self`.

PR #310, opened 2019-11-07
[DRAFT] fix(ffi): return null ptr(s) instead of new objects when applicable

ravinderjangra:
**Needs to test this on all platforms before merge, this PR will generate and upload the libs on S3 which will be used for the testing on all devices**


ravinderjangra:
@bochaco don't merge this one, I am still having some issues with these changes

PR #323, opened 2019-11-16
Implements 'safe cat --hexdump' flag to dump files similar to xxd.

### LINKED ISSUE #189: add a --hexdump arg to cat cmd useful for inspecting binary files
Add an additional arg to the cat command, e.g. --hexdump, to render files them like the output the xxd command (specially useful for binary files):
```
$ cat safe://hnyynys8dhua6yccw5hjdpyjduukrifn94fmb18zkynxr6zyadj1minajrbnc --hexdump
00000000: 4f67 6753 0002 0000 0000 0000 0000 ebba  OggS............
00000010: 756f 0000 0000 c1c9 084b 011e 0176 6f72  uo.......K...vor
00000020: 6269 7300 0000 0002 44ac 0000 0000 0000  bis.....D.......
00000030: 00ee 0200 0000 0000 b801 4f67 6753 0000  ..........OggS..
00000040: 0000 0000 0000 0000 ebba 756f 0100 0000  ..........uo....
00000050: 0553 8299 10f9 ffff ffff ffff ffff ffff  .S..............
00000060: ffff ffff 7103 766f 7262 6973 2c00 0000  ....q.vorbis,...
00000070: 5869 7068 2e4f 7267 206c 6962 566f 7262  Xiph.Org libVorb
00000080: 6973 2049 2032 3031 3530 3130 3520 28e2  is I 20150105 (.
00000090: 9b84 e29b 84e2 9b84 e29b 8429 0900 0000  ...........)....
...
```

dan-da:
Resolves #189.

This PR introduces a new dependency on [pretty-hex](https://crates.io/crates/pretty-hex), with zero sub-dependencies.

Here is example output:

```
$ safe cat --hexdump "safe://hbhyrydxhpah66r4a595drmr5kfxe1d8frp46g3qocg9gswg56ncr4e4im"
Length: 71249 (0x11651) bytes
0000:   7c 44 6f 63  75 6d 65 6e  74 61 74 69  6f 6e 7c 4c   |Documentation|L
0010:   69 6e 75 78  2f 6d 61 63  4f 53 2f 57  69 6e 64 6f   inux/macOS/Windo
0020:   77 73 7c 0a  7c 3a 2d 2d  2d 2d 2d 2d  2d 2d 2d 2d   ws|.|:----------
0030:   2d 3a 7c 3a  2d 2d 2d 2d  2d 2d 2d 2d  2d 2d 2d 2d   -:|:------------
0040:   2d 2d 2d 2d  2d 3a 7c 0a  7c 20 5b 21  5b 44 6f 63   -----:|.| [![Doc
0050:   75 6d 65 6e  74 61 74 69  6f 6e 5d 28  68 74 74 70   umentation](http
0060:   73 3a 2f 2f  64 6f 63 73  2e 72 73 2f  73 61 66 65   s://docs.rs/safe
0070:   2d 63 6c 69  2f 62 61 64  67 65 2e 73  76 67 29 5d   -cli/badge.svg)]
0080:   28 68 74 74  70 73 3a 2f  2f 64 6f 63  73 2e 72 73   (https://docs.rs
<omitted>
```

Notes for discussion or at least to be aware of:

1. An error message is printed by Rust when output is piped and pipe is closed early resulting in SIGPIPE, eg when using 'head'.  example:

```
$ safe cat --hexdump "safe://hbhyrydxhpah66r4a595drmr5kfxe1d8frp46g3qocg9gswg56ncr4e4im" | head -n 4
Length: 71249 (0x11651) bytes
0000:   7c 44 6f 63  75 6d 65 6e  74 61 74 69  6f 6e 7c 4c   |Documentation|L
0010:   69 6e 75 78  2f 6d 61 63  4f 53 2f 57  69 6e 64 6f   inux/macOS/Windo
0020:   77 73 7c 0a  7c 3a 2d 2d  2d 2d 2d 2d  2d 2d 2d 2d   ws|.|:----------
[2019-11-16T19:00:00Z ERROR safe] safe-cli error: Failed to print out the content of the file: Broken pipe (os error 32)
```
This appears to be a [Rust issue](https://github.com/rust-lang/rust/issues/46016).  Still, I don't see it happen with the plaintext file when --hexdump flag is omitted. It may simply be related to length of output buffer though.   It's unclear to me how best to get rid of the error message... open to suggestions.

2. The pretty_hex() func is prepending a "Length" header line.  I think it is fine/nice to have but others might disagree.  Anyway, I [requested a flag](https://github.com/wolandr/pretty-hex/issues/4) in case we decide to remove it.

3. There is no newline at end of dumped output, which results in ugliness when used in a console.  Example:

```
11630:   64 20 74 6f  20 63 6f 6e  74 72 69 62  75 74 65 20   d to contribute 
11640:   74 6f 20 74  68 69 73 20  70 72 6f 6a  65 63 74 2e   to this project.
11650:   0a                                                   .user@host:~/dev/maidsafe/safe-api$
```

I left it this way because normal semantics of 'cat' command is to output exactly the input bytes with no extra newline.  However, a strong argument can be made that in the case of --hexdump, we are already modifying the input bytes so an extra newline to make the output look tidy in a terminal is fine.  I am happy to do it either way though.


dan-da:
oh, and here is how the --hexdump appears in usage.  I didn't bother with a short opt, especially since -h is already taken.

```
FLAGS:
    -n, --dry-run    Dry run of command. No data will be written. No coins spent
    -h, --help       Prints help information
        --hexdump    renders file output as hex
        --json       Sets JSON as output serialisation format (alias of '--output json')
    -V, --version    Prints version information
```

safe-cli/subcommands/cat.rs
```diff
@@ -14,11 +14,15 @@ use prettytable::Table;
 use safe_api::{Safe, SafeData};
 use std::io::{self, Write};
 use structopt::StructOpt;
+use pretty_hex;
 
 #[derive(StructOpt, Debug)]
 pub struct CatCommands {
     /// The safe:// location to retrieve
     location: Option<String>,
+    /// renders file output as hex
```
bochaco: Comestic thing, let's please capitalise first letter

safe-cli/subcommands/cat.rs
```diff
@@ -14,11 +14,15 @@ use prettytable::Table;
...
+    #[structopt(long = "hexdump")]
```
bochaco: How about using `-x` for the short arg? not a big deal though

safe-cli/subcommands/cat.rs
```diff
@@ -65,10 +69,18 @@ pub fn cat_commander(
             }
         }
         SafeData::PublishedImmutableData { data, .. } => {
-            // Render ImmutableData file
-            io::stdout()
-                .write_all(data)
-                .map_err(|err| format!("Failed to print out the content of the file: {}", err))?
+            if cmd.hexdump {
```
bochaco: We can make this a bit more explicit on what's the diff when the condition condition is met with something like:
```
let output = if cmd.hexdump {
   pretty_hex::pretty_hex(data).as_bytes()
} else {
   data
};
io::stdout(output).map_err(...)?
```
So you can clearly see that only the output bytes change, but the rest is all the same including the error we throw if something went wrong.

bochaco:
I think the "Length" header line is ok as you say, we'll see what other think and we can change/remove it if that's what most prefer.
As per the new line at the end, perhaps we can do a println! only when the --hexdump was passed? I see what you mean about cat output, but when the --hexdump is passed I'd compare it to what xxd command does instead of cat.

bochaco:
One last comment I forgot @dan-da , it'd be nice to add one test case for this to https://github.com/maidsafe/safe-api/blob/master/safe-cli/tests/cli_cat.rs, you can run CLI tests using the SCL mock like this:
```
/safe-api$ cd safe-cli
/safe-api$ cargo test --features scl-mock -- --test-threads=1
```
There are some test files at https://github.com/maidsafe/safe-api/tree/master/testdata which are used for automated tests, so you can upload one of those files and then `cat`-it and assert the output?

safe-cli/subcommands/cat.rs
dan-da: Doesn't quite work.  I tried to clone and a couple other things, but no luck.  My rust-foo is weak still.  :-)  Can you suggest a fix?

```
   Compiling safe-cli v0.5.0 (/home/danda/dev/maidsafe/safe-api/safe-cli)
error[E0716]: temporary value dropped while borrowed
  --> safe-cli/subcommands/cat.rs:75:17
   |
73 |             let output = if cmd.hexdump {
   |                 ------ borrow later stored here
74 |                 // hex representation of ImmutableData file
75 |                 pretty_hex::pretty_hex(data).as_bytes()
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary which is freed while still in use
76 |             } else {
   |             - temporary value is freed at the end of this statement
   |
   = note: consider using a `let` binding to create a longer lived value
```

bochaco: yeah that wouldn't be good, there is a way but I'm thinking that if we want to add a new line after the hux dump you can keep it that way and do:
```
use pretty_hex::PrettyHex;
...
if cmd.hexdump {
    // Render hex representation of ImmutableData file
    println!("{:?}", data.hex_dump());
} else {
    io::stdout().write_all(data).map_err(...)?
}    
```

safe-cli/subcommands/cat.rs
dan-da: done.

safe-cli/subcommands/cat.rs
dan-da: done.

safe-cli/subcommands/cat.rs
dan-da: done, using println!

dan-da:
Ok, I added a test case and addressed the other issues you brought up.  I verified that the test case passes (and also fails if I manually change the expected string).

The expect string is a little long so I considered using an md5 digest instead, but decided it was more trouble than its worth.

PR #324, opened 2019-11-18
fix(authd): URL encode passphrase and password when sent to authd

bochaco:
This allows to use any character as part of the passphrase and password when logging-in or creating an account with authd client APIs.

Eventually we'll use JSON-RPC for the messages format rather than passing everything in the URL of the request.


bochaco:
Closing as this will be properly resolved by PR #327

PR #327, opened 2019-11-19
feat(authd): implement json-rpc format for the authd comm

bochaco:


bochaco:
I'll send a PR with that plus some examples of jsonrpc requests/responses

Issue #73, opened 2019-06-13
Commands auto-completion support


bochaco:
Auto completion even coming from SAFE data (for ref: https://www.joshmcguigan.com/blog/shell-completions-pure-rust/)

bochaco:
A first PoC was implemented by community member @draw here: https://safenetforum.org/t/new-release-vault-phase-1-real-vault/29712/145

dan-da:
I've started working on this.  I added a --dump-completions=<shell> flag to safe-cli which dumps shell script that can be sourced by bash, zsh, etc.  This functionality is built into clap, so it automatically uses the values from structopt.  Very elegant.

Activating the completions for bash looks like:

```
safe --dump-completions=bash > /tmp/out && source /tmp/out
```

After that commands, sub-commands and flags auto complete.  slick.  examples:

```
$ safe <tab>
auth                --dry-run           -h                  keypair             -o                  -V
cat                 --dump-completions  help                keys                --output            --version
container           --endpoint          --help              -n                  safe-id             wallet
dog                 files               --json              nrs                 update              --xorurl
```

```
$ safe files 
add                 --endpoint          --help              -o                  sync                --xorurl
--dry-run           -h                  --json              --output            -V                  
--dump-completions  help                -n                  put                 --version
```

```
$ safe cat 
--dry-run           -h                  <location>          --output            --xorurl            
--dump-completions  --help              -n                  -V                  
--endpoint          --json              -o                  --version
```

Ideally at install time,  --dump-completions would be run once for each possible shell and then installed into proper location for user (or system-wide).

Also --dump-completions might be something we only want to build into dev/debug releases.   thoughts advice on that?

bochaco:
That looks really neat, and I agree it's really good if it's generated from the structopt so it can be done with any version of the CLI. I'm wondering if we should have a flag e.g. `--auto-completion` which does the dump and source? Do you know if this will work on Windows too? or if we can implement the Windows equivalent under the same flag?

wardbr:
[...]

PR #328, opened 2019-11-20
Implements --dump-shellcompletions

dan-da:
Addresses #73 for auto-completion of commands in a shell.

Supports shells:
  bash, fish, zsh, powershell, elvish

The completions are automatically generated from the values in
structopt, so they stay current as the cli params change over time.

To use the completions, eg in bash:

    safe --dump-completions=bash > /tmp/out && source /tmp/out

**Considerations**

1. --dump-completions works for me, but could be renamed to anything, eg --completions is shorter.

2. I didn't use a shortopt (yet).   Can add if wanted.

3. All the logic is contained in a single if statement.. I could move it into a function if desired.

4. Nothing is specific to safe-cli app.  If there is a common code shared between the apps under safe-api, then I could make/move a func into the common area to be called by each of them, so each would have --dump-completions.  Each would need a modification to structopt, and a single fn call.

5. Does *not* presently support completion of dynamic text such as a URL that user may be entering.  See discussion in #73.

6. To be most useful, cargo install (or whatever installer) should have a step to dump completions and install for user.   Exactly how best to do that with would need a bit of research, especially when dealing with various shells and OSes.  For now, this PR lays necessary groundwork.

### DIFF
--- a/safe-cli/cli.rs
+++ b/safe-cli/cli.rs
@@ -19,6 +19,7 @@ use crate::subcommands::files::files_commander;
 use crate::subcommands::keys::key_commander;
 use crate::subcommands::networks::networks_commander;
 use crate::subcommands::nrs::nrs_commander;
+use crate::subcommands::setup::setup_commander;
 use crate::subcommands::update::update_commander;
 use crate::subcommands::wallet::wallet_commander;
 use crate::subcommands::{OutputFmt, SubCommands};
@@ -105,6 +106,7 @@ pub fn run_with(cmd_args: &[&str], mut safe: &mut Safe) -> Result<(), String> {
             update_commander().map_err(|err| format!("Error performing update: {}", err))
         }
         Some(SubCommands::Keys(cmd)) => key_commander(cmd, output_fmt, &mut safe),
+        Some(SubCommands::Setup(cmd)) => setup_commander(cmd, output_fmt),
         Some(other) => {
             // We treat these separatelly since we need to connect before
             // handling any of these commands
--- a/safe-cli/subcommands/mod.rs
+++ b/safe-cli/subcommands/mod.rs
@@ -17,6 +17,7 @@ pub mod keys;
 pub mod networks;
 pub mod nrs;
 pub mod safe_id;
+pub mod setup;
 pub mod update;
 pub mod wallet;
 
@@ -83,6 +84,9 @@ pub enum SubCommands {
     #[structopt(name = "files")]
     /// Manage files on the SAFE Network
     Files(files::FilesSubCommands),
+    #[structopt(name = "setup")]
+    /// Perform setup tasks
+    Setup(setup::SetupSubCommands),
     #[structopt(name = "keypair")]
     /// Generate a key pair without creating and/or storing a SafeKey on the network
     Keypair {},
--- a/safe-cli/subcommands/setup.rs
+++ b/safe-cli/subcommands/setup.rs
@@ -1,0 +1,129 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use super::helpers::serialise_output;
+use super::OutputFmt;
+use crate::cli::CmdArgs;
+use std::io::Write;
+use structopt::{clap, StructOpt};
+
+// Defines subcommands of 'setup'
+#[derive(StructOpt, Debug)]
+pub enum SetupSubCommands {
+    /// Dump shell completions.
+    #[structopt(name = "completions")]
+    Completions {
+        /// one of: [bash, fish, zsh, powershell, elvish]  default = all shells
+        shell: Option<clap::Shell>,
+    },
+}
+
+// handles 'setup <cmd>' commands.
+pub fn setup_commander(cmd: SetupSubCommands, output_fmt: OutputFmt) -> Result<(), String> {
+    // Let's keep this clean and place each cmd handler in its own func.
+    match cmd {
+        SetupSubCommands::Completions { shell } => setup_completions(shell, output_fmt),
+    }
+}
+
+// differentiates between 'setup completions' and 'setup completions <shell>'
+fn setup_completions(shell: Option<clap::Shell>, output_fmt: OutputFmt) -> Result<(), String> {
+    match shell {
+        Some(shell_id) => setup_completions_dumpone(shell_id, output_fmt),
+        None => setup_completions_dumpall(output_fmt),
+    }
+}
+
+// handles 'setup completions <shell>' command.  dumps completions for single shell.
+fn setup_completions_dumpone(shell: clap::Shell, output_fmt: OutputFmt) -> Result<(), String> {
+    let buf = gen_completions_for_shell(shell)?;
+
+    if OutputFmt::Pretty == output_fmt {
+        // Pretty format just writes the shell completion to stdout
+        std::io::stdout()
+            .write_all(&buf)
+            .map_err(|err| format!("Failed to print shell completions. {}", err))?;
+        println!();
+    } else {
+        // will be serialized as a string.  no object container.
+        match std::str::from_utf8(&buf) {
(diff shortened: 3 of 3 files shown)


safe-cli/cli.rs
```diff
@@ -68,6 +71,23 @@ pub fn run_with(cmd_args: &[&str], mut safe: &mut Safe) -> Result<(), String> {
         CmdArgs::from_iter_safe(cmd_args).map_err(|err| err.to_string())?
     };
 
+    // If requested, generate shell completions and return
+    if let Some(shell) = args.dump_completions {
+        // Note: here we use name of currently running executable.
+        // Possibly could use value of [[bin]].name in Cargo.toml instead
+        let exec_name = std::env::current_exe()
+            .expect("Can't get the exec path")
```
bochaco: We need to handle the error is as it will panic otherwise, this can be achieved with `std::env::current_exe().map_err(|err| format!("Can't get the exec path: {}", err))?`

safe-cli/cli.rs
```diff
@@ -68,6 +71,23 @@ pub fn run_with(cmd_args: &[&str], mut safe: &mut Safe) -> Result<(), String> {
...
+            .file_name()
+            .expect("Can't get the exec name")
```
bochaco: The error here also needs to handled so it cannot panic.

safe-cli/cli.rs
```diff
@@ -68,6 +71,23 @@ pub fn run_with(cmd_args: &[&str], mut safe: &mut Safe) -> Result<(), String> {
...
+            .to_string_lossy()
+            .into_owned();
+
+        // Generates shell completions for <shell> and prints to stdout
+        CmdArgs::clap().gen_completions_to(exec_name, shell, &mut std::io::stdout());
+
+        return Ok(());
```
bochaco: I think we don't need to return here, it's a flag so it should coexist with other commands and flags. If we don't want it to coexist with other flags then we need it to be a subcommand, or a flag of a subcommand. This is related to a previous comment I made about perhaps thinking of some subcommand which can encapsulate these type of maintenance or setup operations, e.g. `$ safe setup --dump-completions <shell>`.

dan-da: I don't think it makes sense to have it as a flag without returning and then continue on to possibly process other command(s).  eg if I ran safe --dump-completions bash cat <url>, it would prepend before the cat output.

I will see about implementing a setup command.

dan-da: working on setup cmd, but I will just note a counter argument, which is that --help works like --dump-completions in that it returns immediately.  eg *safe --help shell* never enters the shell, so there is precedent.

btw, I am trying to implement it like:

```
safe setup completions dump bash --> bash dump
safe setup completions list --> list of available shells completions.
```

bochaco: The exception proves the rule ;)  seriously, I think --help/-h is definitely the exception and no other case shall exist, I believe, otherwise you may end up in situations where you don't know which flag was applied and which wasn't since one of them was applied and exited.

I think we can handle it with just `$ safe setup completions bash` and if we do something similar to what we did with `--xorurl` flag we can just have `$ safe setup completions --help` to give you what's supported in the help menu. 

We don't have commands or flags for documentation, that's what the help is for, in fact there is a way to have --help to return additional information than -h: https://docs.rs/structopt/0.2.14/structopt/#help-messages, and we were thinking of adding much more help there, even some examples and steps like the user guide but when you do --help.

We should have done the same for the `--output` now that I think of.

dan-da:
Ok, I've rebased and re-worked this quite a lot.  Well, before your last comment I already had safe setup completions <cmd> partially working, so I decided to finish that up first and I added a dumpall cmd.  If nothing else, at least I think this work provides a nice infrastructure in the setup.rs file so that setup commands can easily be 2 levels deep if needed, and for adding other commands.

Regarding the changes you suggested, I'm at a good point now where I could refactor to do it that way, though will likely have to wait until after thxgiving weekend.  Anyway, I wanted to run this implementation by you first.  If you hate the completion <cmd> approach, that's ok.  might be overkill.

One thing I will point out regarding '--json setup completions list' is that the output is machine readable eg by an installer, whereas I think output of --help is not so much.  maybe not a concern, especially if dumpall equivalent remains.

Here's some example usage:

```
$ safe help setup
SUBCOMMANDS:
    completions    Work with shell completions for: [bash, fish, zsh, powershell, elvish]
```

```
$ safe help setup completions
SUBCOMMANDS:
    dump       Dump shell completions.
    dumpall    Dump completions for all shells.
    help       Prints this message or the help of the given subcommand(s)
    list       List shells that have completions available.
```

Each completions sub-command can output either pretty or serialized.  eg:

```
$ safe setup completions list
+------------+
| bash       |
+------------+
| elvish     |
+------------+
| fish       |
+------------+
| powershell |
+------------+
| zsh        |
+------------+
```

```
$ safe setup completions list --json
[
  "bash",
  "elvish",
  "fish",
  "powershell",
  "zsh"
]
```

bochaco:
We don't need to rush it @dan-da , so no worries.

> whereas I think output of --help is not so much. maybe not a concern, especially if dumpall equivalent remains.

I think an install script would go the other way around, it would try to detect which shell is supported in the target OS, and then dump its completion passing the shell as arg, so still not sure we need this type of information about CLI arguments to be returned as part of any command really.

As per dump and dumpall, i would make dumpall to be the default when no shell name is provided as arg to the dump command, e.g.:
`$ safe setup completions dump` => dump all
`$ safe setup completions dump bash` => dump bash only

We could even remove the dump argument, unless we already envision more subcommands of the `completions` subcommand, e.g.:
`$ safe setup completions` => dump all
`$ safe setup completions bash` => dump bash only
I'd vote for this to keep the level of subcommands as shallow as possible.

dan-da:
On Linux at least, a user can have multiple shells installed, though probably not so common and most will just have bash.

In thinking about all this and the big picture, I go back to my original goal, which was just to add a simple, consistent way to dump completions for each of the safe apps, eg: safe-cli, safe-vault, safe-authd, etc.

Looking at safe_vault usage, I notice it does not presently accept any commands, only flags/options.  safe-authd does accept commands. 

Implementing this as a setup sub-command has been a good exercise for me as I learn rust, but I'm not convinced it's the best approach with the above goal in mind.   Would we add 'setup completions' to safe_vault, safe-authd?  Seems kind of heavy for such a simple thing.

So I go back to thinking we should make it either *--completions <shell>* option or *completions <shell>* command and have it work the same in all CLI apps (and maybe gui apps eventually).

There are not very many available shell options and they can be displayed in the help text for the option/command.

what do you think?

bochaco:
yes, if you want to have consistency then perhaps as you say I don't see many more option around this command so simply `$ safe completions` (dump all) and `$ safe completions bash` could do it just fine

dan-da:
ok, I implemented `$ safe completions` (dump all) and `$ safe completions bash`.

If this version is acceptable, I could also see about implementing it for safe_vault and safe_authd.  any issues with that?

```
$ safe setup --help
...
SUBCOMMANDS:
    completions    Dump shell completions.
    help           Prints this message or the help of the given subcommand(s)
```

```
$ safe setup completions --help
safe-setup-completions 0.6.0
bochaco <gabrielviganotti@gmail.com>, Josh Wilson <joshuef@gmail.com>, Calum Craig <calum.craig@maidsafe.net>, Chris
O'Neil <chris.oneil@gmail.com>
Dump shell completions.

USAGE:
    safe setup completions [FLAGS] [OPTIONS] [shell]

FLAGS:
    -n, --dry-run    Dry run of command. No data will be written. No coins spent
    -h, --help       Prints help information
        --json       Sets JSON as output serialisation format (alias of '--output json')
    -V, --version    Prints version information

OPTIONS:
        --endpoint <endpoint>     Endpoint of the Authenticator daemon where to send requests to. If not provided,
                                  https://localhost:33000 is assumed.
    -o, --output <output_fmt>     Output data serialisation: [json, jsoncompact, yaml]
        --xorurl <xorurl_base>    Base encoding to be used for XOR-URLs generated. Currently supported: base32z
                                  (default), base32 and base64

ARGS:
    <shell>    one of: [bash, fish, zsh, powershell, elvish]  default = all shells
```

safe-cli/cli.rs
dan-da: now handles all errors.

safe-cli/cli.rs
dan-da: implemented as `safe setup completions <shell>` per request.

bochaco:
I think it's good for a start, I think we can merge this and see how things evolve in terms of other subcommands of `setup` are required or not, if there are no more subcommands that we can add to the `setup` subcommand we can then flatten it up a bit and move `completions` to be a direct subcommand of safe.

PR #330, opened 2019-11-20
chore(ci): github actions for release process

jacderida:
This transfers the build and release process from Jenkins to GitHub Actions. The entire process is replicated here, split into 2 workflows, PR and master.

The PR configuration will run the tests and build the mobile configurations. The master configuration will build everything, then deploy and publish. There are conditionals that check for the existence
of a version change message, as before. Since you have the same configuration for building mobile in both workflows, there's a lot of duplication here, but for now we're just accepting that as a limitation
of the Actions system.

This [release](https://github.com/jacderida/safe-cli/releases/tag/0.5.3) on my fork was done using Actions.

Due to an issue with rust_sodium, we are also missing macOS for the time being, but sodium should be removed fairly soon. I've left stuff commented, which you should be able to uncomment when macOS is ready again.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,3 +1,5 @@
+name: Master
+
 on:
   push:
     branches:
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,0 +1,366 @@
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build-component:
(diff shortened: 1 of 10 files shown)


jacderida:
> Can you make a new commit incoporating these changes as well?
> 
> [maidsafe/safe_client_libs#1078](https://github.com/maidsafe/safe_client_libs/pull/1078)

This has now been added.

PR #338, opened 2019-11-25
chore(ci): add cli integration tests with authd and local vault to the GH Actions matrix

### LINKED ISSUE #317: Run automated e2e tests in CI with safe-authd and local vault
Run CLI integration tests in CI using safe-authd and local vault

bochaco:
Resolves #317


bochaco:
Most of this is now being submitted as part of PR #356 . I'll be sending a small separate PR to just enable the tests with local vault whenever the issue in vault is solved.

PR #344, opened 2019-11-29
 feat(cli): support for having config settings and a command to switch networks

### LINKED ISSUE #233: Easily switch connections between local and shared vault
With the recent release of the safe clis with real vaults, along with MaidSafe providing a shared vault that people can connect to, there have been a number of comments/support issues where people have not realised that their `vault_connection_info.config` file for the shared vault has been overwritten when they have launched a local vault.

There have been a number of internal discussions on the best way to handle this and the general consensus now is that this could be addressed within the cli, potentially by being able to pull the latest shared vault config via the cli.

User scenarios which must be taken into account:
1. A user who uses a local vault only. They restart their local vault vault multiple times a day and require the vault_connection_info.config to be overwritten with the latest keys from their local vault.
2. A user who uses a shared vault only. User connects to the shared vault periodically and has never run a vault locally. This user may or may not keep up to date with the latest forum discussions, i.e. may not be aware if we were to announce we had restarted the shared vault and therefore it had a new `vault_connection_info.config` file that they would require.
3. A user who flicks between using a local vault and the shared vault.

bochaco:
Resolves #233


safe-cli/README.md
```diff
@@ -987,6 +976,78 @@ Version: 3
 
 In this case we don't only get information about the content that the URL resolves to, but also about the NRS Map Container this NRS-URL was resolved with. E.g. we see the XOR-URL of the NRS Map Container, its version, and among other data we also see the list of all NRS names defined by it with their corresponding XOR-URL links.
 
+### Networks
+
+The CLI, like any other SAFE application, can connect to different SAFE networks that may be available. As the project advance several networks may coexist apart from the main SAFE Network, there could be SAFE Networks available for testing upcoming features, or networks that are local to the user in their own computer or WAN/LAN.
```
joshuef: `as the project advanceS`

safe-cli/README.md
```diff
@@ -987,6 +976,78 @@ Version: 3
...
+
+Currently, with current state of the project there is a shared Vault which acts as public test network accessible to anyone. And also users may have local vaults running in their own environment. The CLI allows users to easily create a list of different SAFE networks in its config settings, to then be able to switch between them with just a simple command. Let's see some examples.
```
joshuef: `Currently, with current state of the project there is a shared Vault` -> `Currently there is a shared Vault`

bochaco:
@joshuef I just pushed a commit fixing the issues you reported, as well as having a way to cache current network conn info into CLI networks config.
I also added add/remove subcommands for the networks command as it's more convenient, but kept the ones within the config command as this commands should grow and allow to manage all configs in a raw form let's say.

safe-cli/README.md
```diff
@@ -156,21 +159,9 @@ Sending request to authd to obtain an status report...
...
-```shell
-$ mkdir ~/safe_vault
-$ cd ~/safe_vault
-$ wget https://github.com/maidsafe/safe_vault/releases/download/0.19.2/safe_vault-0.19.2-x86_64-unknown-linux-musl.tar.gz
-$ tar -xzvf safe_vault-0.19.2-x86_64-unknown-linux-musl.tar.gz
-$ ./safe_vault
-```
-
-Once we made sure we have a vault running locally (or otherwise the right configuration for a remote vault), we can proceed with the next steps.
+Once you made sure we have a vault running locally or set up the system for a remote network, you can proceed with the next steps.
```
S-Coyle: Could maybe reword this to "Once you have made sure you have a vault running locally, or have the connection file for a remote vault/network in place, you can proceed with the next steps."

safe-cli/README.md
```diff
@@ -987,6 +978,87 @@ Version: 3
...
+Example command to perform this on Linux or Mac:
+```shell
+$ mkdir ~/safe_vault
+$ cd ~/safe_vault
+$ wget https://github.com/maidsafe/safe_vault/releases/download/0.19.2/safe_vault-0.19.2-x86_64-unknown-linux-musl.tar.gz
+$ tar -xzvf safe_vault-0.19.2-x86_64-unknown-linux-musl.tar.gz
+$ ./safe_vault
+```
+
+Once the local vault is running, the connection configuration file will be already in the correct place for your applications (including the CLI) to connect to this vault, so you can simply use the CLI or any application from now on to connect to your local vault. Note that depending on the application, you may need to restart it so it uses the new connection information for your local vault rather than a previously existing one.
```
S-Coyle: Can you switch "be already" to "already be" here

safe-cli/README.md
```diff
@@ -987,6 +978,87 @@ Version: 3
...
+
+#### Switch networks
+
+MaidSafe hosts a vault for those who don't want to run a local vault but still have a go at using the CLI and clients applications. It's very common for users testing and experimenting with CLI and SAFE applications to have a local vault running, but switching to use the MaidSafe shared vault, back and forth, is also quite common.
```
S-Coyle: Should be "client applications." (not "clients")

safe-cli/README.md
```diff
@@ -987,6 +978,87 @@ Version: 3
...
++--------------+----------------------------------------------------------------------------------------------+
+| Network name | Connection info location                                                                     |
++--------------+----------------------------------------------------------------------------------------------+
+| my-vault     | ~/.config/safe-cli/networks/my-vault_vault_connection_info.config                            |
++--------------+----------------------------------------------------------------------------------------------+
+| shared-vault | https://github.com/maidsafe/safe_vault/releases/download/0.20.0/vault_connection_info.config |
++--------------+----------------------------------------------------------------------------------------------+
+```
+
+Once we have them in the CLI setting, we can use the CLI to automatically fetch the connection information data using the configured location, and place it in the right place in the system for SAFE applications to connect to the selected network. E.g. let's switch to the 'shared-vault' network we previously configured:
```
S-Coyle: Should be "CLI settings" (not "setting")

safe-cli/README.md
```diff
@@ -987,6 +978,87 @@ Version: 3
...
+```shell
+$ safe networks switch shared-vault
+Switching to 'shared-vault' network...
+Fetching 'shared-vault' network connection information from 'https://github.com/maidsafe/safe_vault/releases/download/0.20.0/vault_connection_info.config' ...
+Successfully switched to 'shared-vault' network in your system!
+You'll need to login again, and re-authorise the CLI, if you need write access to the 'shared-vault' network
+```
+
+Remember that every time you run a local vault the connection configuration in your system is automatically overwritten by the local vault with new connection information. Also if the shared vault was restarted by MaidSafe, the new connection information is published in the same URL and needs to be updated in your system to be able to successfully connect to it. Thus if you want to make sure your currently setup network matches any of those set up in the CLI config, you can use the `check` subcommand:
```
S-Coyle: "...the new connection information is published in the same URL..." - not got an issue with this phrase, just something that we'll need to keep in mind if we commit to it - every time the shared vault updates we need to copy the config file to all previous shared vault releases as the user may have set up to point to an old release, as per your example. Wonder if we should consider pointing users elsewhere to get this file, maybe S3?

bochaco: right, I think we need another place where we categorise them by "network" rather than by vault version, right now we have only one network/shared-vault but eventually we may have more than one, e.g. shared-vault and fleming-nodes. Unless we plan to do this now we can keep this sentence as is and update it later when we move to that.

PR #351, opened 2019-12-05
feat(connect): use connect API for unregistered client using unregistered auth response

ravinderjangra:


safe-api/src/api/safe_client_libs.rs
```diff
@@ -97,7 +97,14 @@ impl SafeApp for SafeAppScl {
         let app = match auth_credentials {
             Some(auth_credentials) => {
                 let auth_granted = decode_ipc_msg(auth_credentials)?;
-                App::registered(app_id.to_string(), auth_granted, disconnect_cb)
+                match auth_granted {
```
bochaco: We could perhaps prevent from creating this new `AuthResponseType` enum and have the decode_ipc_msg to already return the `App`? so it could be renamed to something like `create_app_from_credentials` and it does it all that's being done here?

PR #352, opened 2019-12-05
chore(ci): update GHA

m-cat:
+ Remove some unneeded stuff
+ Add iOS builds back into PR workflow

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -5,6 +5,11 @@ on:
     branches:
       - master
 
+env:
+  # Run all cargo commands with --verbose.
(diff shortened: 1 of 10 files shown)


m-cat:
Once https://github.com/maidsafe/safe_client_libs/pull/1103 is merged and we do a version update, we'll be able to remove rust_sodium in this PR.

m-cat:
I've updated to latest SCL in this PR and also:

+ added caching (should result in similar benefits to [SCL](https://github.com/maidsafe/safe_client_libs/pull/1117))
+ fixed automatic publishing based on problems we ran into in SCL

PR #356, opened 2019-12-10
feat(cli): connect with read-only access when no credentials were found

### LINKED ISSUE #355: Support read-only access from CLI with all commands to public data

### LINKED ISSUE #359: Cached network configs not removed when setting is deleted from config

### LINKED ISSUE #360: Upgrade SCL to latest version which removes libsodium dependency

### LINKED ISSUE #362: Release scripts to remove mock versions from release notes and deliverables

bochaco:
Resolves #355 
Fixes #359 
Resolves #360 
Resolves #362 

This PR also re-enables CLI e2e tests in CI using authd with mock-network.

The e2e tests with authd and local vault is also part of this PR although they are not run yet as there is an issue presumably in vault (https://github.com/maidsafe/safe-api/issues/317) preventing them from working. They can be enabled once the issue in vault is solved.


.github/workflows/master.yml
```diff
@@ -408,13 +396,6 @@ jobs:
           asset_name: safe-cli-${{ steps.versioning.outputs.cli_version }}-x86_64-apple-darwin.tar.gz
           asset_content_type: application/zip
         if: startsWith(steps.commit_message.outputs.commit_message, 'Version change')
-      - uses: actions/upload-release-asset@v1.0.1
-        with:
-          upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: resources/safe_completion.sh
```
joshuef: we're getting rid of the completion?

safe-api/Cargo.toml
```diff
@@ -33,10 +33,10 @@ rand = "0.6.5"
 rand_core = "0.4.0"
 rcgen = "0.2"
 relative-path = "0.4.0"
-safe_app = "0.11.0"
-safe_authenticator = "0.11.0"
-safe_core = "0.36.0"
-safe-nd = "~0.4.0"
+safe_app = { git = "https://github.com/maidsafe/safe_client_libs.git", branch = "master"}
```
joshuef: we're still waiting on SCL release I guess?

.github/workflows/master.yml
bochaco: We now have a CLI command for it (implemented by @dan-da ), so we'll need to just add a section in the user guide, we can do that in a separate PR. So we don't need the static script in the repo and/or in the release notes.

safe-api/Cargo.toml
bochaco: Unless we are creating a release right now, until then, we can use latest commit to keep up with SCL changes?

PR #364, opened 2019-12-12
Show QR code from the CLI to scan and connect with mobile devices

ravinderjangra:
targeted on #356 

gen qr code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1819,6 +1819,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "qrcodegen"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "quic-p2p"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2376,6 +2381,7 @@ dependencies = [
  "pretty-hex 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "pretty_assertions 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "prettytable-rs 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "qrcodegen 1.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "relative-path 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "reqwest 0.9.24 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


safe-ffi/ffi/mod.rs
```diff
@@ -23,8 +23,35 @@ use errors::Result;
...
+    ),
+) {
+    catch_unwind_cb(user_data, o_cb, || -> Result<()> {
+        let user_data = OpaqueCtx(user_data);
+        let app_id = String::clone_from_repr_c(app_id)?;
+        let app_name = String::clone_from_repr_c(app_name)?;
+        let app_vendor = String::clone_from_repr_c(app_vendor)?;
+        let endpoint = from_c_str_to_str_option(endpoint);
+        let auth_response = Safe::auth_app(&app_id, &app_name, &app_vendor, endpoint)?;
+        let auth_response_c_str = CString::new(auth_response)?;
```
m-cat: This variable name is a bit misleading as it makes me think of `CStr` which is a different type than `CString`.

Also, in general I would avoid encoding the type of a variable into its name, it is easily missed in a refactor and also just unnecessary. I would just shadow this over the previous variable and call it `auth_response`.

resources/test-scripts/all-tests
```diff
@@ -2,7 +2,6 @@
 
 set -e -x
 
-source ./resources/test-scripts/run-auth-daemon
```
m-cat: I don't know the background of these changes but maybe you want to run the new e2e test scripts here?

resources/get_release_description.sh
```diff
@@ -53,29 +40,39 @@ zip: ZIP_WIN_CHECKSUM
...
 ```
 
-## FFI
+### SAFE Authenticator daemon
+
+The Authenticator daemon exposes services which allow applications and users to create SAFE Network accounts, log in using an existing account's credentials (passphrase and password), authorise applications which need to store data on the network on behalf of the user, as well as revoke permissions previously granted to applications.
+The SAFE Authenticator, which runs as a daemon or as a service in Windows platforms, can be started and managed with the SAFE CLI if the `safe-authd`/`safe-authd.exe` binary is in the system's PATH with execution permissions (please refer to [Authenticator section in CLI User Guide](https://github.com/maidsafe/safe-api/blob/master/safe-cli/README.md#the-authenticator-daemon-authd) for detailed instructions).
+
+| OS | Download link | SHA-256 checksum |
+| --- | --- |
```
m-cat: Looks like there are only two columns here when there should be three

resources/test-scripts/all-tests
ravinderjangra: The changes are not in this PR. The PR is based on #356

resources/get_release_description.sh
ravinderjangra: The changes are not in this PR. The PR is based on #356

safe-ffi/ffi/mod.rs
ravinderjangra: Will update this.

bochaco:
I'l pushed a tiny commit to PR #356 covering the issues you raised @m-cat , thanks!

bochaco:
I think we should try to (not necessarily in this PR) render the QR code with safe logo, e.g. https://bochaco.github.io/safe-qrcode-svg/

PR #374, opened 2019-12-16
docs(UserGuide): add text to files sync section covering `--update-nrs` flag

S-Coyle:


safe-cli/README.md
```diff
@@ -684,6 +684,14 @@ FilesContainer synced up (version 2): "safe://hbyw8kkqr3tcwfqiiqh4qeaehzr1e9boiu
 
 The `./other-folder/file1.txt` file will be uploaded and published in the `FilesContainer` with path `/new-files/file1.txt`.
 
+One more thing to note about Files Sync is the use of the `--update-nrs` flag. If you want to sync the NRS URL rather than the XOR-URL (see NRS section below for more information about NRS Containers and commands) then this must be specified using this flag:
```
bochaco: The update-nrs flag updates the NRS name **in addition to** syncing the FilesContainer at the XOR-URL you provide and bumping its version.

safe-cli/README.md
```diff
@@ -684,6 +684,14 @@ FilesContainer synced up (version 2): "safe://hbyw8kkqr3tcwfqiiqh4qeaehzr1e9boiu
...
+```shell
+$ safe files sync ./to-upload/ safe://mywebsite --update-NRS
```
bochaco: The flag should be lower case, I thought it was cosmetic but I jsut tried and CLI doesn't like it it's not lower case, I guess since we could have upper case args like `-V` vs. `-v`

PR #376, opened 2019-12-16
doc(user-guide): instructions for installing shell completions

dan-da:
Some basic instructions for installing shell completions for bash and powershell.   

The powershell instructions are untested as I don't have a windows install handy right now.  Would be great if someone can try/improve the powershell instructions.

Also, I stuck these instructions under "Using the CLI" to minimize changes but it might make sense to create a top-level "Installation" section and put it under there.


bochaco:
I can try the windows on a VM in the next couple of days. I was thinking this should go to the bottom of the content under "Using the CLI" but beside the "Update" cmd/section? I just think this is not the first thing you are interested in knowing about the SAFE CLI but an "advanced" feature, I'm specially thinking in terms of not putting off newbies who want to connect-to/try SAFE as quickly and with the fewest steps and reading as possible.

PR #378, opened 2019-12-17
chore(CI): build android libs without make and docker

ravinderjangra:


m-cat:
Cool, I'll take a look later. Is it possible to remove the docker file also? Would be nice to remove the makefile as well, we can move the remaining jobs into separate scripts in another PR.

bochaco:
The builds for ffi android never start, so not sure if the conditionals are not working there @ravinderjangra ?

ravinderjangra:
@bochaco if you check the list you will find the different set of Android checks because the names have been changed.

PR #379, opened 2019-12-17
chore(CI): simplify iOS build using cargo-lipo

ravinderjangra:

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -69,19 +69,25 @@ jobs:
 
   build-ffi-ios:
     name: Build FFI iOS
-    runs-on: macOS-latest
-    strategy:
-      matrix:
-        target: [aarch64-apple-ios, x86_64-apple-ios]
+    runs-on: macos-latest
+    env:
+      PKG_CONFIG_ALLOW_CROSS: 1
+      DEVELOPER_DIR: /Applications/Xcode_11.2.app/Contents/Developer
     steps:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
         with:
-          profile: minimal
           toolchain: stable
+          profile: minimal
           override: true
-          target: ${{ matrix.target }}
-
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -99,48 +105,29 @@ jobs:
...
-          command: build
-          args: --release --manifest-path=safe-ffi/Cargo.toml --target=${{ matrix.target }} --features=mock-network
-      # Upload for build-ffi-ios-universal
+          name: safe-ffi-apple-ios-prod
+          path: target/universal/release/libsafe_ffi.a
+      # Build dev universal library.
+      - name: Build universal lib - Dev
+        shell: bash
+        run: cargo lipo --release --manifest-path=safe-ffi/Cargo.toml --features=mock-network
+      # Upload prod universal library.
```
m-cat: This comment should say "dev"

.github/workflows/pr.yml
```diff
@@ -69,19 +69,25 @@ jobs:
...
+          override: true
+          target: aarch64-apple-ios
+      - uses: actions-rs/toolchain@v1
+        with:
           toolchain: stable
+          profile: minimal
           override: true
-          target: ${{ matrix.target }}
-
+          target: x86_64-apple-ios
```
m-cat: Just a question, did you mean to use two toolchains in the same job? I'm a bit confused by this, maybe because I don't understand lipo and universal builds, but some comments here would be nice.

m-cat: Looking at the logs, it looks like both toolchains are installed successfully but only one of them is set as active:

https://github.com/maidsafe/safe-api/pull/379/checks?check_run_id=352743435

ravinderjangra: hey @m-cat, yes we need both the toolchains because we are generating universal libs (aarch64 + x86_64). previously those were done separately and then we were generating fat/universal lib which is now done using cargo-lipo.

PR #383, opened 2020-01-09
feat(authd): use S3 as backend server for installing and updating safe-authd binary

### LINKED ISSUE #361: Enhance and simplify authd installation, as well as support authd upgrade from cli

bochaco:
Resolves #363 
Resolves #361


bochaco:
This depends on a new version available of self-update crate. A PR was sent to self-update (https://github.com/jaemk/self_update/pull/35) with implementation to support S3, once that's merged and released we can then merge this PR.

PR #388, opened 2020-01-20
refactor(authd): migrate to use async/await

### LINKED ISSUE #366: Move quic server impl from authd crate onto jsonrpc-quic crate

bochaco:
Resolves #366 

- Refactor jsonrpc-quic crate to expose JSON-RPC over QUIC client and server endpoint logic without a runtime
- Refactor jsonrpc-quic to expose JSON-RPC requests and responses objects and hide serialisation/deserialisation logic behind its API
- Refactor safe-authd and safe-api to remove quic server and client logic, to make use jsonrpc-quic client/server endpoint
- Refactor safe-authd to start using Rust async/await
- Upgrade futures to v0.3.4 in jsonrpc-quic and safe-authd
- Upgrade tokio to v0.2.11 in all crates
- Upgrade quinn to v0.5.3, safe-nd to v0.8.0, self_update to v0.13.0
- Add missing commands to the interactive shell


safe-api/src/api/helpers.rs
```diff
@@ -125,7 +124,7 @@ pub fn parse_coins_amount(amount_str: &str) -> Result<Coins> {
         match err {
             SafeNdError::ExcessiveValue => Error::InvalidAmount(format!(
                 "Invalid safecoins amount '{}', it exceeds the maximum possible value '{}'",
-                amount_str, MAX_COINS_VALUE
+                amount_str, Coins::from_nano(4_294_967_295_999_999_999)
```
joshuef: If this value needs to be in `Coins` could that not then be changed in `MAX_COINS_VALUE` const? I worry about deviating from using the same const...

bochaco: @joshuef that const was removed from safe-nd, that's why I changed that, if that's what you mean?

PR #395, opened 2020-01-21
feat(cli): add files ls command

### LINKED ISSUE #365: Addition of `safe files ls` command
Based on discussions on the forum (summarised here: https://forum.safedev.org/t/file-uid-as-sha256sum/2666/18) add the `safe files ls` command which creates an abstraction on the FilesContainers exposing the same options/flags as a traditional ls command would do on any file system, allowing to see which are files and which are subfolders, and also resolving paths and showing the content of only that path and not the entire FilesContainer (note this is already implemented in the cat command, so we can replicate that behaviour or move it and only make it possible with files ls).

bochaco:
Resolves #365


joshuef:
Would be good to get a test cast for this into cli integration too :+1:

PR #415, opened 2020-01-30
style(typos): fix some typos noticed in a few files

S-Coyle:


S-Coyle:
cargo fmt failed in CI - latest rustc v1.41.0  had a new fmt check so have added a commit to address that

S-Coyle:
New clippy also threw up some issues - commit pushed to address those

PR #420, opened 2020-02-12
feat(rustls): Use self-udpate fork w/ rustls-tls flag set for reqwest.

joshuef:
Temp change to use rustls for self_update via a fork, until we know how/if this is wanted upstream


bochaco:
Closing as these changes have been already made as part of PR #388 .

PR #424, opened 2020-02-19
feat(files): use file's xorurl to decide if it needs to be re-uploaded in a files sync operation

### LINKED ISSUE #271: `safe files sync` fails to update changed files
I uploaded an app to `safe://solid-filemanager` which worked, made some changes and rebuilt the ./build directory and then attempted to update the shared vault as follows. But I see that while new files have been uploaded and deleted files are deleted, the app does not work because although `./build/index.html` has changed it was not updated, which causes the app to fail as it tries to load the files which have been deleted. You can see from the output below that `index.html` is not update:
```
safe files sync -r ./build/ safe://solid-filemanager -u -d
FilesContainer synced up (version 1): "safe://solid-filemanager"
+  ./build/precache-manifest.af230ca88da8d9c886cd5120bbb0119a.js  safe://hbyyyydxwtwoox6s8i6kt5cdfw3ne1adqj3gyb5dq81h91k6uq19krkm43 
+  ./build/static/js/main.2c523573.chunk.js                       safe://hbyyyyn387fq7nacoya9kanp659nf9895ogd3gw34d1gxq35r4pci8rtci 
+  ./build/static/js/main.2c523573.chunk.js.map                   safe://hbyyyyd8mm1s645mys3mqdcyjzpobzwyawsaoqtsz6wqnozogptjat79tj 
-  /precache-manifest.c959321ddd7159002846116ea7c06a1b.js         safe://hbyyyydoxadadgm6cagec68gfduioti1r1m5ctcoi6trsrnofhjbn5tcc5 
-  /static/js/main.59a99085.chunk.js                              safe://hbyyyynmor3r8qiwdtafiggp5i3t7bw4iewmxehsw1fnsf3udwtg7iko17 
-  /static/js/main.59a99085.chunk.js.map                          safe://hbyyyyduim63o4jgcme7nycxxxpyr1dqa1cxpegazaxg1eoczbsb7wc8x9
```
Inspecting the container with `safe cat safe://solid-filemanager?v=0` (and v=1) you can confirm the contents have only been partially updated.

## Your Environment

I'm using the shared vault with:

- safe_auth: 0.1.0
- safe: 0.2.2
- SAFE Browser 0.15.1
- Operating System and version: Ubuntu 19.04
- Link to your project: https://github.com/theWebalyst/solid-filemanager/tree/safe-network-able

Note: you should not try to reproduce this with my project as it depends on two other modules that I've forked, but I think you should be able to reproduce it using a similar project with the SAFE CLI commands in the terminal.

bochaco:
Resolves #271

### DIFF
--- a/safe-api/src/api/files.rs
+++ b/safe-api/src/api/files.rs
@@ -751,8 +751,7 @@ fn files_map_sync(
             Some(file_item) => {
                 if force
                     || (compare_file_content
-                        && (file_item[FAKE_RDF_PREDICATE_SIZE] != file_size
-                            || file_item[FAKE_RDF_PREDICATE_TYPE] != file_type))
+                        && is_file_modified(safe, &Path::new(local_file_name), file_item))
                 {
                     // We need to update the current FileItem
                     if add_or_update_file_item(
@@ -814,6 +813,13 @@ fn files_map_sync(
     });
 
     Ok((processed_files, updated_files_map, success_count))
+}
+
+fn is_file_modified(safe: &mut Safe, local_filename: &Path, file_item: &FileItem) -> bool {
+    match upload_file_to_net(safe, local_filename, true /* dry-run */) {
+        Ok(local_xorurl) => file_item[FAKE_RDF_PREDICATE_LINK] != local_xorurl,
+        Err(_err) => false,
+    }
 }
 
 fn files_map_add_link(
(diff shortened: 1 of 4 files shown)


safe-api/src/api/files.rs
```diff
@@ -2140,9 +2194,9 @@ mod tests {
         assert_eq!(new_processed_files.len(), 1);
         assert_eq!(new_files_map.len(), 2);
         assert_eq!(
-        new_processed_files["../testdata/test.md"].1,
-        "File named \"/sub2.md\" already exists on target. Use the \'force\' flag to replace it"
-    );
+            new_processed_files["../testdata/test.md"].1,
+            "File named \"/sub2.md\" already exists on target. Use the \'force\' flag to replace it"
```
joshuef: We should probably distinguish no? It exists, is different to exists and _is the same_.

bochaco: yeah it would be nice for UX, let me try to do that, I think it should be possible now that we have the xorurls

PR #425, opened 2020-02-24
feat(api): add range based request for the immutable data

ravinderjangra:
depends on https://github.com/maidsafe/safe_client_libs/pull/1169

### DIFF
--- a/safe-api/src/api/app/fetch.rs
+++ b/safe-api/src/api/app/fetch.rs
@@ -234,7 +234,7 @@ fn fetch_from_url(safe: &Safe, url: &str, retrieve_data: bool, range: Range) -> 
                         let new_target_xorurl = file_item.get("link")
                             .ok_or_else(|| Error::ContentError(format!("FileItem is corrupt. It is missing a \"link\" property at path, \"{}\" on the FilesContainer at: {} ", path, xorurl)))?;
 
-                        safe.fetch(new_target_xorurl, None)
+                        safe.fetch(new_target_xorurl, range)
                     }
                     None => Err(Error::ContentError(format!(
                         "No content found matching the \"{}\" path on the FilesContainer at: {} ",
@@ -278,7 +278,7 @@ fn fetch_from_url(safe: &Safe, url: &str, retrieve_data: bool, range: Range) -> 
             debug!("Resolving target from resolvable map: {}", url_with_path);
 
             let (_, public_name, _, _) = get_subnames_host_path_and_version(url)?;
-            let content = safe.fetch(&url_with_path, None)?;
+            let content = safe.fetch(&url_with_path, range)?;
             the_xor.set_path(""); // we don't want the path, just the NRS Map xorurl and version
             let nrs_map_container = NrsMapContainerInfo {
                 public_name,
@@ -363,6 +363,8 @@ mod tests {
     use crate::api::xorurl::XorUrlEncoder;
     use rand::distributions::Alphanumeric;
     use rand::{thread_rng, Rng};
+    use std::fs::File;
+    use std::io::Read;
     use unwrap::unwrap;
 
     #[test]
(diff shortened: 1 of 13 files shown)


ravinderjangra:
Do we need some test cases for this?

safe-api/src/api/fake_scl.rs
```diff
@@ -255,7 +255,12 @@ impl SafeApp for SafeAppFake {
         Ok(xorname)
     }
 
-    fn files_get_published_immutable(&self, xorname: XorName) -> Result<Vec<u8>> {
+    fn files_get_published_immutable(
+        &self,
+        xorname: XorName,
+        _position: Option<u64>,
+        _length: Option<u64>,
```
bochaco: I second @lionel1704 comments about adding tests for this feature, just using some of the files we have in `testdata` folder as we do for the rest of the tests

ravinderjangra:
> I'm wondering that probably the most common case is to not pass any range and therefore combining these two values in a tuple could be better, e.g. range: Option<(Option<u64>, Option<u64>)>, we may even want to declare that as a type in the API

I think having a new type for this seems like overkill but if that's what we want then I can update the PR for the same. 

> I think we may want to perhaps expose a new `fetch_range` API

You mean a new `files_get_published_immutable_range` API? cause the `fetch_range` doesn't make more sense cause it returns different types of data and may cause confusion.

bochaco:
> cause the `fetch_range` doesn't make more sense cause it returns different types of data and may cause confusion.

I'm assuming thins like the browser use the `fetch`, so I thought having a `fetch_range` is what they would instead use if they see a rage in the request coming from the app, I presume one of the main use cases here is streaming from the web apps

ravinderjangra:
> > cause the `fetch_range` doesn't make more sense cause it returns different types of data and may cause confusion.
> 
> I'm assuming things like the browser use the `fetch`, so I thought having a `fetch_range` is what they would instead use if they see a range in the http request coming from the app, I presume one of the main use cases here is streaming from the web apps

If you are fine then I think I will go with the `files_get_published_immutable_range` and for the stream related apps, since we know it's a specific can we can use this API. I'll try this or maybe I'll add both for more flexibility.

bochaco:
I think we need @joshuef input here to confirm what would be needed for the Browser to be able to support streaming

joshuef:
Yo, just catching up.

-----
> I'm assuming things like the browser use the fetch

That's something that's been brought up, and I'm thinking maybe calling these apis `fetch` is misleadingly `HTTPy`... As we're not returning headers compatible for this. So maybe just `get` data as that's more generic.

------
As for a separate API... I think in terms of immutable, it's overkill... `Option<Option<RangeStart>,Option<RangeEnd> >` could be fine. It could/would also work for `Sequence` of data, or range of versions and such things... 

Options as you could pass only one, and get from that byte, or to that end-byte...

safe-api/src/api/fetch.rs
```diff
@@ -131,11 +133,16 @@ impl Safe {
     /// assert!(data_string.starts_with("hello tests!"));
     /// ```
     pub fn inspect(&self, url: &str) -> Result<SafeData> {
-        fetch_from_url(self, url, false)
+        fetch_from_url(self, url, false, None)
     }
 }
 
-pub fn fetch_from_url(safe: &Safe, url: &str, retrieve_data: bool) -> Result<SafeData> {
+pub fn fetch_from_url(
```
bochaco: This should private

safe-api/src/api/safe_client_libs.rs
```diff
@@ -256,14 +257,31 @@ impl SafeApp for SafeAppScl {
...
 
-    fn files_get_published_immutable(&self, xorname: XorName) -> Result<Vec<u8>> {
+    fn files_get_published_immutable(&self, xorname: XorName, range: Range) -> Result<Vec<u8>> {
         debug!("Fetching immutable data: {:?}", &xorname);
 
         let safe_app: &App = self.get_safe_app()?;
         let immd_data_addr = IDataAddress::Pub(xorname);
         let data = run(safe_app, move |client, _app_context| {
-            immutable_data::get_value(client, immd_data_addr, /*decryption_key:*/ None)
+            if range.is_some() {
```
bochaco: I think here you should either use a `match range {` or `if let Some((offset, len)) = range {`

safe-ffi/ffi/constants.rs
```diff
@@ -0,0 +1,14 @@
...
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// http://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+/// Read entire contents of a file.
+pub static FILE_READ_TO_END: u64 = 0;
```
bochaco: Shouldn't they be `pub const` ?

safe-api/src/api/fetch.rs
ravinderjangra: This should be private. I was also little confused when I was adding the range here. I'll update this one.

safe-ffi/ffi/constants.rs
ravinderjangra: It doesn't make much difference since not used by any lang code so I'll update this one.

safe-api/src/api/safe_client_libs.rs
ravinderjangra: I wasn't sure about the best way here so I'll try to use what you suggested and update this.

safe-ffi/ffi/constants.rs
bochaco: I think the difference is in terms what this value really is, the semantic when you see it, I assumed it was a constant which won't change, thus `const`, if the plan is that it's a global variable which may eventually change then that's what `static` should be used for, this is from https://doc.rust-lang.org/rust-by-example/custom_types/constants.html

safe-api/src/api/app/fetch.rs
```diff
@@ -232,7 +234,7 @@ pub fn fetch_from_url(safe: &Safe, url: &str, retrieve_data: bool) -> Result<Saf
                         let new_target_xorurl = file_item.get("link")
                             .ok_or_else(|| Error::ContentError(format!("FileItem is corrupt. It is missing a \"link\" property at path, \"{}\" on the FilesContainer at: {} ", path, xorurl)))?;
 
-                        safe.fetch(new_target_xorurl)
+                        safe.fetch(new_target_xorurl, None)
```
bochaco: Shouldn't you be passing the `range` here?

safe-api/src/api/app/fetch.rs
```diff
@@ -276,7 +278,7 @@ pub fn fetch_from_url(safe: &Safe, url: &str, retrieve_data: bool) -> Result<Saf
             debug!("Resolving target from resolvable map: {}", url_with_path);
 
             let (_, public_name, _, _) = get_subnames_host_path_and_version(url)?;
-            let content = safe.fetch(&url_with_path)?;
+            let content = safe.fetch(&url_with_path, None)?;
```
bochaco: Shouldn't you be passing the `range` here?

safe-ffi/ffi/fetch.rs
```diff
@@ -33,7 +36,18 @@ pub unsafe extern "C" fn fetch(
...
     catch_unwind_cb(user_data, o_err, || -> Result<()> {
         let url = String::clone_from_repr_c(url)?;
-        let content = (*app).fetch(&url);
+        let pos = if index == FILE_READ_FROM_START {
+            None
+        } else {
+            Some(index)
+        };
+
+        let len = if length == FILE_READ_TO_END {
```
bochaco: The `pos` and `len` variables should be renamed to `end` as per what the values are expected to be, i.e. the start and end positions

safe-ffi/ffi/files.rs
```diff
@@ -229,7 +232,19 @@ pub unsafe extern "C" fn files_get_published_immutable(
...
         let user_data = OpaqueCtx(user_data);
         let url_str = String::clone_from_repr_c(url)?;
-        let data = (*app).files_get_published_immutable(&url_str)?;
+        let pos = if index == FILE_READ_FROM_START {
+            None
+        } else {
+            Some(index)
+        };
+
+        let len = if length == FILE_READ_TO_END {
```
bochaco: Same as above about the names of `pos` and `len` variables

safe-ffi/ffi/mod.rs
```diff
@@ -9,6 +9,7 @@
 
 #![allow(clippy::missing_safety_doc)]
 
+pub mod constants;
```
bochaco: This should be a private mod I think

safe-api/src/api/app/fetch.rs
ravinderjangra: I wasn't sure cause the range will be only used for the immutable data for now. but it seems there may be recursive fetch calls so I'll just update this now.

safe-ffi/ffi/fetch.rs
ravinderjangra: That should be fine. I'll update the code to use start and end instead of pos and len. I'll update the API logic as well to get the length using end-start.

ravinderjangra:
@bochaco, I'm seeing one test failure after rebasing on the latest master. Could you please check this one?

safe-api/src/api/app/fetch.rs
bochaco: I fixed some tests which were failing for the same reason, I'm not sure why we didn't see this one though in previous CI runs, it's anyway due to a change in test files which is now detected with a different mime type, so we can just change the test case assertion there.

bochaco:
I fixed some tests which were failing for the same reason, I'm not sure why we didn't see this one though in previous CI runs, it's anyway due to a change in test files which is now detected with a different mime type, so we can just change the test case assertion there.

PR #481, opened 2020-03-26
`safe files get`

dan-da:
This PR implements `safe files get`.   Basic usage is:

`safe files get <source>  [dest]`

The command will copy all files from source to dest, which defaults to '.'

If files(s) already exist in dest, the default action is to *ask* the user whether to overwrite for each file.  However the --exists flag allows user to specify *overwrite* all or *preserve* all.  It's a bit wordy having to type out "overwrite" or "ignore", but is very clear.  Could be argued that we should have some single-char equivalent for each.

A challenge we face is that the file system data (directories, files) is opaque to the server (vaults), so if a client wants only a subset of directories or files from a FileContainer, then the full list must be obtained from the server and then filtered by the client before downloading the desired subset of files.  For example:

`safe files get safe://asdfasdfdafadsfdsaf/testdata/subfolder /tmp`

A user would expect the above command to copy subfolder to /tmp/subfolder, which is what happens.

I plan that a future PR will enhance the filtering functions to support shell expansion chars, such as wildcards: * and ?.    sets:  {1,2,3}    ranges  {4-6}

I believe that these sorts of filtering operations will be a common requirement of SAFE applications beyond just the CLI.  As such, I made a first stab at implementing APIs in safe-api to facilitate such.

Also, the `get` operation should work both for files stored as either public or private immutable data.   As such, I created an API that will fetch either, depending on type encoded in the xorurl.  The private data getter is presently unimplemented() but I wanted to get the abstraction in place at least.

This is a big PR and I don't expect it to be accepted right away.  I have more tests to write, some of the APIs may need to be tweaked or maybe won't be accepted as APIs.   

But I do do feel the code is at a state where it is ready to start getting some review/feedback.

Notes:
* support for empty directories is planned, but requires modification to `safe files put` first.
* support for preserving file attributes is planned, but requires modification to `safe files put` first.
* support for preserving symlinks is planned, but requires modification to `safe files put` first.
* There are a couple minor-ish issues with the indicatif progress bars.  I've opened issues in the indacitif repo, so hopefully resolved soon.


dan-da:
So I am a bit conflicted about whether the functions I put into safe-api should be APIs or not, at this point in development.

My initial reasoning is that a big part of SAFE is to be a networked filesystem.  As such, many clients will need to perform file/directory retrievals and we should facilitate that in a straight-forward way.  Path and wildcard filtering/expansion is especially tricky, so if we provide a standard implementation, it will make the system much more approachable for developers, and also apps will tend to behave more similarly for end-users.

That said, the present APIs may be a bit too focused on the use-case for `safe files get` and too high level, compared to other APIs which are pretty much network primitives.  I still think SAFE should provide these functionalities, but perhaps safe-api/src/api/app/files.rs is the wrong place to do it, and more thought should be given to the design/interface.

As such, I think for the short term, I will move all code into safe-cli/subcommands/files_get.rs, and leave the API untouched.   That will get us the `files get` functionality quickly, and we can discuss/refactor path filtering and other possible APIs later.

bochaco:
@dan-da the first comment I was gonna make in this PR was about the FilesMap functions, analogous to the NrsMap struct we have where all the resolution logic resides, we should functions which work on the FilesMap struct, this was part of the original plan but we never did it, you can see a TODO comment in this other place which has some logic that needs to be moved out to a FilesMap struct impl: https://github.com/maidsafe/safe-api/blob/master/safe-api/src/api/app/fetch.rs#L225-L226

In this way, just like NrsMap APIs, files APIs return a FilesMap, the FilesMap object hides the logic for the data representation (RDF, etc.) and it also exposes the utilities to search/filter files, again hiding the data representation logic when dealing with the search/filtering.

dan-da:
I refactored all the code out of safe-api and into safe-cli/subcommands/files_get.rs.  So at this time, there are no changes to the safe-api.  

@bochaco, given present time constraints and other priorities, I figured it is most practical to just get `safe files get` working well/correctly as a self-contained thing.  We can then refactor relevant portions of it into FilesMap API in a more generalized fashion.  Anyway, the wildcard matching and range/set expansion (ala bash) is still TODO, and would be a big part of that.

Further, I made significant changes/fixes to the way that source and dest paths are handled.  With these changes, it correctly handles get'ing a single file specified in url path as well as a container (no url path) or subfolder, and also renaming.   In other words, it "just works" as anyone familiar with `cp` or `scp` would expect.

I also created 15+ integration tests.

Once this PR is merged, I view the following as my/our remaining TODOs to have a complete upload/download solution:
* `safe files put` uploads hidden files/dirs, and empty dirs (tagged via metadata)
* `safe files put` includes files metadata such ctime, mtime, is_symlink and perm bits similar to what git does.
* `safe files get` is able to read the above metadata and (optionally) write out, eg with --archive flag.
* `safe files get` performs filtering of wildcards and range/sets expansion.

dan-da:
Latest:
* found/fixed a regression when saving to relative paths and '.'
* added a test case for above
* added usage/examples in safe-cli/README.md

dan-da:
rebased to master.

PR #498, opened 2020-03-31
chore/ffi: add a dummy extn fn to prevent safe_auth_ffi symbols from …

ravinderjangra:
…being removed


bochaco:
@ravinderjangra are the commit and PR comment correct? they don't seem to relate to the actual change?

PR #506, opened 2020-04-07
expose authenticator FFI

ravinderjangra:


safe-ffi/ffi/authenticator/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+                };
+                o_err(user_data.0, &ffi_result.into_repr_c()?);
+            }
+        }
+        Ok(())
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn autherise_unregistered_app(
```
bochaco, line 278: Typo in "autherise_unregistered_app"

What is this function for? we don't authorise unregistered apps as they only get the connection info in a response from such a requests, right?

ravinderjangra: ya, it does not authorize anything but only returning the serialized connection info as response.

PR #523, opened 2020-04-16
feat/api: parse query params in any type of URL and keep them as part of XorUrlEncoder info

bochaco:


safe-api/src/api/app/xorurl.rs
```diff
@@ -419,12 +430,18 @@ impl XorUrlEncoder {
...
 
-        match self.content_version {
-            Some(v) => Ok(format!("{}?v={}", xorurl, v)),
-            None => Ok(xorurl),
+        xorurl = match self.content_version {
+            Some(v) => format!("{}?v={}", xorurl, v),
+            None => xorurl,
+        };
+
+        for q in self.query_params.iter() {
```
dan-da, line 440: when self.content_version is None, this will result in a url with "?" before the query params.

bochaco: ah yes, i just forced-push a fix for that

safe-api/src/api/app/xorurl.rs
```diff
@@ -206,6 +210,7 @@ impl XorUrlEncoder {
         path: Option<&str>,
         sub_names: Option<Vec<String>>,
         content_version: Option<u64>,
+        query_params: Vec<String>,
```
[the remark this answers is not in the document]
bochaco: Yes, lets go with a single string, same for fragment, I'll make that change tomorrow

PR #525, opened 2020-04-16
feat/api: make the `inspect` API to return the complete URL resolution chain

bochaco:
Also, adapt the dog command to show every step in the resolution chain, and resolves #518 
Example `dog` command output with a long URL resolution chain:
```
$ safe dog safe://subfile

== URL resolution step 1 ==
= NRS Map Container =
PublicName: "subfile"
XOR-URL: safe://subfile
Version: 0
Type tag: 1500
XOR name: 0xaaebb5968b8eabdda8f9185cfca60198861b12a370b63bc5799edd7ce0bca2c3
Native data type: PublishedSeqAppendOnlyData
+------------------+----------------------+----------------------+------------------------------+
| NRS name/subname | Created              | Modified             | Link                         |
+------------------+----------------------+----------------------+------------------------------+
| subfile          | 2020-05-06T04:38:30Z | 2020-05-06T04:38:30Z | safe://subfolder/sub2.md?v=0 |
+------------------+----------------------+----------------------+------------------------------+

== URL resolution step 2 ==
Resolved from: safe://subfolder/sub2.md?v=0
= NRS Map Container =
PublicName: "subfolder"
XOR-URL: safe://subfolder?v=0
Version: 0
Type tag: 1500
XOR name: 0x505528ad158a19d4cd9360bcb644798100d30e432f206ddaff5607021477a986
Native data type: PublishedSeqAppendOnlyData
+------------------+----------------------+----------------------+-------------------------------+
| NRS name/subname | Created              | Modified             | Link                          |
+------------------+----------------------+----------------------+-------------------------------+
| subfolder        | 2020-05-06T04:38:08Z | 2020-05-06T04:38:08Z | safe://testdata/subfolder?v=0 |
+------------------+----------------------+----------------------+-------------------------------+

== URL resolution step 3 ==
Resolved from: safe://testdata/subfolder?v=0
= NRS Map Container =
PublicName: "testdata"
XOR-URL: safe://testdata?v=0
Version: 0
Type tag: 1500
XOR name: 0xebd25cfe070ab282250533a201e38c83249d489a3bf1c8f9718bad6369f59994
Native data type: PublishedSeqAppendOnlyData
+------------------+----------------------+----------------------+----------------------------+
| NRS name/subname | Created              | Modified             | Link                       |
+------------------+----------------------+----------------------+----------------------------+
| testdata         | 2020-05-06T04:37:50Z | 2020-05-06T04:37:50Z | safe://myname/testdata?v=0 |
[...]

PR #532, opened 2020-05-04
fix(cli): adds inspect_to_xorurl() and resolve_to_xorurl() APIs

dan-da:
This PR has code to recursively resolve an NRS-URL chain to a final XOR-URL.  It works similarly to Fetch::retrieve_from_url() but in a simplified fashion that focuses only on (NRS) url resolution, so it is easier to use and to audit/review for that purpose.  The PR includes a test case for recursive URL resolution in the CLI.  The two new APIs are:  Safe::inspect_to_xorurl() and Safe::resolve_to_xorurl() in fetch.rs.

Now, at the same time, there is #525 that improves/fixes recursive resolution for `cat` and `dog` commands.  In theory, those funcs should be able to do it, but in practice I encountered difficulties adapting for use in Safe::parse_and_resolve_to_url().  (@bochaco maybe you can do it?)   So this PR may be redundant, at least in the core APIs.     Regardless, there are some useful mods in the CLI and a test case.

As its unclear if these new APIs will be merged/used, I have not provided doctests.  Also, possibly/probably they should return XorUrlEncoder (SafeUrl) instead of a String url, to save callers needing to parse in some instances.  If this PR seems likely to get merged, I can make those changes.

Summary of changes:
* adds Safe::inspect_to_xorurl() and Safe::resolve_to_xorurl() APIs
* refactor Safe::parse_and_resolve_url() to use Safe::resolve_to_xorurl()
* modify 'files tree' and 'files ls' to fully resolve nrs urls
* adds a test for recursive nrs resolution in cli_files_get

### DIFF
--- a/safe-api/src/api/app/fetch.rs
+++ b/safe-api/src/api/app/fetch.rs
@@ -140,6 +140,17 @@ impl Safe {
     pub async fn inspect(&self, url: &str) -> Result<SafeData> {
         retrieve_from_url(self, url, false, None).await
     }
+
+    /// # Resolves a safe:// URL, which may be a chain of NRS urls, to a final XOR-URL.
+    pub async fn resolve_to_xorurl(&self, url: &str) -> Result<String> {
+        resolve_to_xorurl(self, url).await
+    }
+
+    /// # Inspects a safe:// URL, which may be a chain of NRS urls, and returns all links
+    ///   in the chain along with associated xorurl's.
+    pub async fn inspect_to_xorurl(&self, url: &str) -> Result<Vec<(String, Option<String>)>> {
+        inspect_to_xorurl(self, url).await
+    }
 }
 
 async fn retrieve_from_url(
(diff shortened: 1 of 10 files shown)


bochaco:
>  but in practice I encountered difficulties adapting for use in Safe::parse_and_resolve_to_url(). (@bochaco maybe you can do it?)

Yes, it's actually already done and working here: https://github.com/maidsafe/safe-api/pull/525/files#diff-6ad5f8dfdeda7eb56cb924eb5727fcbaR71, I'll be finalising that PR once PR #529 is merged

dan-da:
> Yes, it's actually already done and working here: https://github.com/maidsafe/safe-api/pull/525/files#diff-6ad5f8dfdeda7eb56cb924eb5727fcbaR71, I'll be finalising that PR once PR #529 is merged

Ok, sounds good.  You might want to try out the test case `files_get_src_is_nrs_recursive_and_dest_not_existing` from this PR with your implementation.

bochaco:
FYI, I'm working on fixing the scenario for `files_get_src_is_nrs_recursive_and_dest_not_existing` along with a refactoring which removes some redundant logic. This is in WIP PR #535, it's currently based on a non-merged PR, so the changes I've worked out so far are in the last commit: https://github.com/maidsafe/safe-api/pull/535/commits/94079528f3dc7b5cf488fcf4a964b633cde60d1d

PR #535, opened 2020-05-06
fix/cli: files get was not resolving URLs with more than one resolution step and paths

bochaco:
A refactoring is also made here to:
- Remove URL resolution redundant logic from CLI in favor of using the `fetch` API
- Remove the `parse_and_resolve_url` from being part of the public API and keeping it only as an internal helper function
- Disabling CI tests with mini-network as there is a known issue in SCL which causes them to fail

### DIFF
--- a/safe-api/src/api/app/nrs.rs
+++ b/safe-api/src/api/app/nrs.rs
@@ -40,7 +40,8 @@ impl Safe {
     // Parses a safe:// URL and returns all the info in a XorUrlEncoder instance.
     // It also returns a second XorUrlEncoder if the URL was resolved from an NRS-URL,
     // this second XorUrlEncoder instance contains the information of the parsed NRS-URL.
-    pub async fn parse_and_resolve_url(
+    // *Note* this is not part of the public API, but an internal helper function used by API impl.
+    pub(crate) async fn parse_and_resolve_url(
         &self,
         url: &str,
     ) -> Result<(XorUrlEncoder, Option<XorUrlEncoder>)> {
--- a/safe-cli/subcommands/files.rs
+++ b/safe-cli/subcommands/files.rs
@@ -8,9 +8,7 @@
 // Software.
 
 use super::{
-    files_get::{
-        filter_files_map_by_xorurl_path, process_get_command, FileExistsAction, ProgressIndicator,
-    },
+    files_get::{process_get_command, FileExistsAction, ProgressIndicator},
     helpers::{
         gen_processed_files_table, get_from_arg_or_stdin, get_from_stdin, if_tty, notice_dry_run,
         parse_stdin_arg, pluralize, serialise_output,
@@ -18,15 +16,16 @@ use super::{
     OutputFmt,
 };
 use ansi_term::Colour;
+use log::debug;
 use prettytable::{format::FormatBuilder, Table};
 use safe_api::{
+    fetch::SafeData,
     files::{FilesMap, ProcessedFiles},
     xorurl::{XorUrl, XorUrlEncoder},
     Safe,
 };
 use serde::Serialize;
-use std::collections::BTreeMap;
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use structopt::StructOpt;
 
 type FileDetails = BTreeMap<String, String>;
@@ -375,32 +374,18 @@ async fn process_tree_command(
 ) -> Result<(), String> {
     let target_url = get_from_arg_or_stdin(target, Some("...awaiting target URl from STDIN"))?;
 
-    let (_version, files_map) = safe
-        .files_container_get(&target_url)
-        .await
-        .map_err(|err| format!("Make sure the URL targets a FilesContainer.\n{}", err))?;
-
-    let filtered_filesmap =
-        filter_files_map_by_xorurl_path(&files_map, &target_url, |urlpath, fmpath| {
-            // remove urlpath from translated path
-            // eg, urlpath:  /project/src/module
-            //      fmpath:  /project/src/module/submod/foo.rs
-            //  translated:  /submod/foo.rs
-            if !urlpath.is_empty() {
-                let mut newpath = fmpath.to_string();
-                newpath.replace_range(..urlpath.len(), "");
-                Some(newpath)
-            } else {
-                None
-            }
-        })?;
+    debug!("Getting files in container {:?}", target_url);
+    let files_map = match safe.fetch(&target_url, None).await? {
+        SafeData::FilesContainer { files_map, .. } => files_map,
+        _other_type => return Err("Make sure the URL targets a FilesContainer.".to_string()),
+    };
 
     // Create a top/root node representing `target_url`.
     let mut top = FileTreeNode::new(&target_url, FileTreeNodeType::Directory, Option::None);
     // Transform flat list in `files_map` to a hierarchy in `top`
     let mut files: u64 = 0;
     let mut dirs: u64 = 0;
-    for (name, file_details) in filtered_filesmap.iter() {
+    for (name, file_details) in files_map.iter() {
         let path_parts: Vec<String> = name
             .to_string()
             .trim_matches('/')
--- a/safe-cli/subcommands/files_get.rs
+++ b/safe-cli/subcommands/files_get.rs
@@ -16,6 +16,7 @@ use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget, ProgressStyle, T
 use log::{debug, info, trace, warn};
 use safe_api::{
     fetch::Range,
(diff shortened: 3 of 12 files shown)


safe-cli/subcommands/files_get.rs
```diff
@@ -407,19 +405,6 @@ fn prompt_yes_no(prompt_msg: &str, default: &str) -> bool {
     }
 }
 
-/// # Downloads files within a FilesContainer that match the xorurl path component and writes them to disk, preserving paths.
-///
-/// TODO: In the future, this will have options for preserving symlinks and
-/// file attributes.
-/// ```
-async fn files_container_get_matching(safe: &Safe, url: &str) -> ApiResult<(u64, FilesMap)> {
```
dan-da, line 415: I'm not understanding yet why this needed to go away.   Maybe I need to take a step back and look at the changes as a whole, but i was just expecting that Safe::parse_and_resolve_url() would return a fully resolved URL, which would require only minimal changes in this file.

safe-cli/subcommands/files_get.rs
```diff
@@ -667,71 +659,6 @@ async fn files_map_get_files(
     Ok(processed_files)
 }
 
-/// # Filter out file items outside of xorurl path
```
dan-da, line 670: Umm, so where did this filtering go?   Did it move into FilesMap or something?

safe-cli/subcommands/files.rs
```diff
@@ -375,32 +374,18 @@ async fn process_tree_command(
...
-            if !urlpath.is_empty() {
-                let mut newpath = fmpath.to_string();
-                newpath.replace_range(..urlpath.len(), "");
-                Some(newpath)
-            } else {
-                None
-            }
-        })?;
+    debug!("Getting files in container {:?}", target_url);
+    let files_map = match safe.fetch(&target_url, None).await? {
```
dan-da, line 378: hmm, you removed the path translation above.   Does that mean that safe.fetch() is translating the paths inside the FileContainer before returning?

safe-api/src/api/app/nrs.rs
```diff
@@ -38,23 +38,49 @@ impl Safe {
...
-        if xorurl_encoder.content_type() == SafeContentType::NrsMapContainer {
-            let (_version, nrs_map) = self.nrs_map_container_get(&url).await.map_err(|_| {
-                Error::InvalidInput(
-                    "The location couldn't be resolved from the NRS URL provided".to_string(),
-                )
-            })?;
-            let xorurl = nrs_map.resolve_for_subnames(xorurl_encoder.sub_names_vec())?;
-            Ok((XorUrlEncoder::from_url(&xorurl)?, Some(xorurl_encoder)))
-        } else {
+        // We don't want to resolve the path, so let's make sure it's
```
dan-da, line 47: I don't quite understand this statement "We don't want to resolve the path".  Why not?    Perhaps you can comment it in a way that makes it clear how recursive path resolution works, in context of:
* https://github.com/maidsafe/rfcs/blob/master/text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md#3-path-and-version-resolution and
* https://github.com/maidsafe/rfcs/blob/master/text/0052-RDF-for-public-name-resolution/0052-RDF-for-public-name-resolution.md#5-recursive-nrs-url-resolution

bochaco:
> Does that mean that safe.fetch() is translating the paths inside the FileContainer before returning?

Yes, that's the answer to almost everything. The `fecth` API is **the resolver**
, and it resolves paths too. If you look at the `cat` command you will see it calls `fetch` only that's why it works with all the scenarios we've been talking about with paths and multiple indirection hops.

Regarding the `parse_and_resolve` function, I want to get rid of it, or at least remove it from being a public API and perhaps just a helper for only a couple of places where it may still be needed.

I'll be trying to finalise this PR in the next couple of days, but this PR already has most of tests passing as is, I still have to go through five/six `cli_files_get` tests which are failing. API tests are all passing already.

In summary, I think the overall is like this:
- `fetch` is the resolver, for any type of content and type of resolution supported by the network, e.g. NRS, FilesContainers, Wallet paths, and whatever we can have in the future.

- All the get APIs (e.g. files_container_get, nrs_map_container_get, etc.) resolve only XOR/NRs URLs, but only the version, nothing else. These APIs are used if you want to retrieve a specific type of content, and you want the data, like a FilesMap or NRSMap. These functions make use of `parse_and_resolve`, which is why it doesn't resolve paths, and that's why I also think it doesn't make any more sense to expose it as a public API since we have the `fetch`.

`parse_and_resolve` is even only needed in a few places, but most of the other places can also use `fetch` internally I think, which is also something I'd like to eventually do.

safe-cli/subcommands/files_get.rs
```diff
@@ -430,48 +415,60 @@ async fn files_container_get_files(
...
-    // Else we just use the xor_url path.
-    let resolved_url_encoder = match nrs_encoder {
-        Some(p) => {
-            let mut s = xorurl_encoder.path_decoded()?;
-            s.push_str(&p.path_decoded()?);
-            let mut e = xorurl_encoder.clone();
-            e.set_path(&s);
-            e
+    debug!("Getting files in container {:?}", url);
+    let (version, files_map, xorurl_encoder) = match safe.fetch(url, None).await? {
```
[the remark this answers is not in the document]
bochaco: That's what e want, and `fetch` is the resolver.

> I thought you were going to modify Safe::parse_and_resolve_url() to return the fully resolved URL.

That's done already, but in the other PR, PR #525

safe-api/src/api/app/nrs.rs
```diff
@@ -40,7 +40,8 @@ impl Safe {
     // Parses a safe:// URL and returns all the info in a XorUrlEncoder instance.
     // It also returns a second XorUrlEncoder if the URL was resolved from an NRS-URL,
     // this second XorUrlEncoder instance contains the information of the parsed NRS-URL.
-    pub async fn parse_and_resolve_url(
```
dan-da, line 43: Why is this being removed from public API?

I think there is a lot of value in being able to call a function that simply resolves an arbitrary SafeUrl to its final form (XorUrl).  Basically the SAFE equivalent of gethostbyname() in C, which is available in pretty much every programming language.

yes, the caller can call the fetch functions directly, but if all they want is the final xorurl, there is fairly complex/verbose logic to perform on the result, eg, the body of parse_and_resolve_url().

bochaco: As said above, parse_and_resolve doesn't solve paths, it cannot resolve paths because that's how it's used as a helper by other functions, i.e. to not resolve paths. So a caller cannot use it as a resolver, it can use `fetch` as also point out, and if the caller is interested in a XorUrlEncoder/SafeUrl instance then all that's needed is `let safe_url = SafeUrl::from_url(safe.fetch(<url>)?.xorurl())?;`

dan-da: > let safe_url = SafeUrl::from_url(safe.fetch(<url>)?.xorurl())?;

ah ok, nice.  easier than I thought, due to SafeData::xorurl().   But perhaps still not totally obvious.  I'd feel better if we add an API for it.  Perhaps something along the lines of:

```
impl SafeUrl {
    pub fn resolve_to_xorurl(&self) -> Result<SafeUrl>{
        if self.is_nrs_url() {
            SafeUrl::from_url(safe.fetch(url)?.xorurl())?;
        }
        else {
            self.clone();   // any way to avoid this clone?
                            // could throw an error instead -- force caller to check is_xorurl() first.
        }
    }
}
```

just an idea, does't have to be part of this PR if you don't like it.

bochaco: We could, but we could also could have many more helpers functions, I think we should expose those which are strictly necessary at least for now to reduce the dev and maintenance cost on the bindings (nodejs, C#, etc.). 
Why would you have the `if self.is_nrs_url()` condition there?

dan-da: >  I think we should expose those which are strictly necessary

fair enough.  your call.

> Why would you have the if self.is_nrs_url() condition there?

Because if the self SafeUrl is an nrsurl then we need to resolve it, but if it is an xorurl then it is already final, so no resolution needed.  At least, that's how I see it.

bochaco: We can have this discussion later if we go for such an API, but a SafeUrl can be a XOR-URL and still not being resolved, e.g. I could construct a SafeUrl using the XOR-URL of an NRS Map container.

dan-da: ah yes, then perhaps the proposed API should be named differently, eg just SafeUrl::resolve().

PR #551, opened 2020-05-20
feat(cli): put/sync includes "hidden" files, and omit sub-dirs if not recursive.

dan-da:
Changes in this PR:

1. removes the filter for files/dirs beginning with `.` when performing PUT or SYNC
2. omits sub-directories from PUT or SYNC if recursive flag not present.  (prevents empty sub-dirs)
3. updates tests to expect the above changes.

Rationale for 1:

* The check for files beginning with `.` is a platform specific (unix) method to identify hidden files and filter them out.  However, hidden files work differently on windows, so arguably this check wasn't even accomplishing its purpose on that platform, and possibly others.
* As far as I can see, there is no valid reason to filter hidden files in mandatory fashion.  It would definitely surprise users that are familiar with cp, scp, or rsync, as these tools do not do this.
* Decisions about this kind of filtering should be made by the individual app/dev, not by safe-api.
* If we decide to implement such an option to omit hidden files in end-user tool(s), such as safe-cli, then probably we should expose a filtering callback, or alternative mechanism for app to build FilesMap itself.  Anyway, that could be a separate PR.

Rationale for 2:
* Previously, a non-recursive PUT would include sub-dirs without their children. (empty)  That's somewhat weird, as it is kinda/sorta like recursion, but not.
* Act like familiar tools. `cp testdata/* /tmp/testdata/` omits child directories.  (it does emit a warning, perhaps API should log one)

Note that the `testdata/emptyfolder` directory is not actually empty because it contains a `.gitkeep` file.  Git doesn't actually support empty directories, so we can't keep a truly empty dir in a git repo.  Since `files put` is now including `testdata/emptyfolder/.gitkeep` in the FilesContainer, some tests that were depending on emptyfolder being empty broke.  To remedy this, I made a function in the tests to create a truly empty folder for each test as such: `<sys_tmp_dir>/<random_name>/emptyfolder`.  The `random_name` parent dir gets deleted after the test.

### DIFF
--- a/safe-api/src/api/app/files.rs
+++ b/safe-api/src/api/app/files.rs
@@ -1401,14 +1401,14 @@ async fn file_system_dir_walk(
     info!("Reading files from {}", file_path.display());
     let (metadata, _) = get_metadata(&file_path)?;
     if metadata.is_dir() || !recursive {
-        // TODO: option to enable following symlinks and hidden files?
+        // TODO: option to enable following symlinks?
         // We now compare both FilesMaps to upload the missing files
         let max_depth = if recursive { MAX_RECURSIVE_DEPTH } else { 1 };
         let mut processed_files = BTreeMap::new();
         let children_to_process = WalkDir::new(file_path)
             .follow_links(true)
             .into_iter()
-            .filter_entry(|e| not_hidden_and_valid_depth(e, max_depth))
+            .filter_entry(|e| valid_depth(e, max_depth))
             .filter_map(|v| v.ok());
 
         for (idx, child) in children_to_process.enumerate() {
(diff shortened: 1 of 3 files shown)


bochaco:
> Previously, a non-recursive PUT would include sub-dirs without their children

Can you give an example of this, unless I misunderstand, this wasn't the case, e.g. with safe v0.12.0:
```
safe files put ../testdata/ 
FilesContainer created at: "safe://hnyynyigkyxyhmpgke7n1secrpef57o4rrrbwqk17go53oukdqket8zfhgbnc"
+  ../testdata/another.md   safe://hbhynydt5b95dmumcm8yig4u1keuuh8hgsr5yx39xn4mqikp91sbdhbpwp 
+  ../testdata/noextension  safe://hbyyyynw7fi7ko8u3wtai6eekko3m5y6u9ud7w1i4ajycmwfwu3wwpnn81 
+  ../testdata/test.md      safe://hbhynydpan7d94mwp1bun3mxfnrfrui131an7ihu11wsn8dkr8odab9qwn 
```

dan-da:
> > Previously, a non-recursive PUT would include sub-dirs without their children
> 
> Can you give an example of this, unless I misunderstand, this wasn't the case, e.g. with safe v0.12.0:

The issue was introduced by my last PR for empty folder support, where PUT now understands each dir as a "thing" distinct from its children.  Sorry if the word *Previously* was unclear.

Issue #510, opened 2020-04-09
nrs paths ignored by 'safe files ls' and 'safe files tree'


dan-da:
1. PUT a folder with subfolders.
```
$ safe files put ../testdata --recursive
FilesContainer created at: "safe://hnyynysryo3qp5s481fp9srtpk5bj643kebrcxiy9tuz8a4kugnfbk9fxabnc"
+  ../testdata/another.md              safe://hbhynydjthmxbrth4tkgrjki8mcihd5frigrz4yunw4rsjh31qhnaombg9 
+  ../testdata/noextension             safe://hbyyyydj99a5hm68szc3ehnt4swbo7caziod9w53b5j87efsykjqyxoo4k 
+  ../testdata/subfolder/sub2.md       safe://hbhynynt4oa3rihpop4sjppp3j5y59aa93a9ca7mqdkghg3mo9s76wgnis 
+  ../testdata/subfolder/subexists.md  safe://hbhynyn9dgh89nhe4o5sb4hs8n8csynq8qrjckf4nt8aqmkwa9uebhargk 
+  ../testdata/test.md                 safe://hbhynydjpor31taroabod6fugmbk5h9i4t7zz1edzu9oohrour9jogank8
```

2. create an nrs link to a subfolder
```
$ safe nrs create sf --link safe://hnyynysryo3qp5s481fp9srtpk5bj643kebrcxiy9tuz8a4kugnfbk9fxabnc/testdata/subfolder?v=0
New NRS Map for "safe://sf" created at: "safe://hnyydyi6mr4ruodhqoj1q9jerqhawnass3ormbdteusecf8a7nkdz6h5qebqh"
+  sf  safe://hnyynysryo3qp5s481fp9srtpk5bj643kebrcxiy9tuz8a4kugnfbk9fxabnc/testdata/subfolder?v=0
```

3. observe that `safe cat` correctly displays contents of the subfolder
```
$ safe cat safe://sf
Files of FilesContainer (version 0) at "safe://sf":
+--------------+------+----------------------+----------------------+-------------------------------------------------------------------+
| Name         | Size | Created              | Modified             | Link                                                              |
+--------------+------+----------------------+----------------------+-------------------------------------------------------------------+
| sub2.md      | 4    | 2020-04-09T17:35:57Z | 2020-04-09T17:35:57Z | safe://hbhynynt4oa3rihpop4sjppp3j5y59aa93a9ca7mqdkghg3mo9s76wgnis |
[...]

PR #558, opened 2020-05-21
fix(cli): addresses issue 510. 'files ls' now uses nrs paths

dan-da:
This PR contains a fix for #510, specifically for `files ls`, as `files tree` was already addressed.

Changes:
* `files ls` now uses Safe::inspect() api to resolve URLs.
* added optimization to filter_files_map() for single-file case, which can do a simple path lookup.
* added/moved/changed test cases:
    - calling_files_ls_on_nrs_with_path()
    - calling_files_ls_on_single_file
    - calling_files_ls_with_invalid_path()
* adds safe_cmd() to safe-cmd-test-utilities, a succinct way to exec, check exit code, and get all command output.

@bochaco please have a look at the changes in files.rs for `Ls` subcommand.  I had a lot of difficulty trying to get it to work for a single file path using the Safe::fetch() API, so I came up with an alternate solution using Safe::inspect().  I commented what it is doing carefully, to make the approach clear.  Anyway, I'm hoping you can come up with something cleaner, and/or modify the fetch() API somehow to deal with this type of case, eg maybe have an option to fetch until FileContainer found, and no further.

### DIFF
--- a/safe-api/src/api/app/fetch.rs
+++ b/safe-api/src/api/app/fetch.rs
@@ -19,6 +19,7 @@ pub use super::{
 use crate::{Error, Result};
 use log::{debug, info};
 use serde::{Deserialize, Serialize};
+use std::path::Path;
 
 pub type Range = Option<(Option<u64>, Option<u64>)>;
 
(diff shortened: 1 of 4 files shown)


bochaco:
> eg maybe have an option to fetch until FileContainer found

Ok, I'll take a look at it tomorrow, if that's all we need it's straight forward

bochaco:
I think we should be doing this: https://github.com/dan-da/safe-api/pull/2, you'll just need to fix the total-files number reported, and you don't need the `filter_files_map` function.

bochaco:
I just realised that https://github.com/dan-da/safe-api/pull/2 won't report the correct version of the FilesContainer, so this other reports it correctly: https://github.com/dan-da/safe-api/pull/3

dan-da:
ok, all checks have passed, just needs a review.

PR #569, opened 2020-06-05
chore(nlt): update network launch tool

ravinderjangra:


ravinderjangra:
We re-open this once updated to use the latest scl and safe-nd.

PR #572, opened 2020-06-08
fix/cli: show the XOR-URL of an NRS Container (as part of dog cmd output) without any subname

bochaco:
It resolves the issue reported on the SAFE forum: https://safenetforum.org/t/xor-url-from-dog-xor/31885


joshuef:
only thought is to maybe add CLI test for this?

bochaco:
This fix i now part of PR #583

PR #573, opened 2020-06-09
Test safe-cli against phase-2b vaults

ravinderjangra:


bochaco:
I think we need to still keep test with baby fleming disabled, the problem will be solved once we have CRDT hopefully

ravinderjangra:
I think we are fine with disabling the tests though don't we want to ensure the cli and APIs are working fine with the latest vaults? Because I noticed some test failures.

bochaco:
L which wasn't solved yet, and we hope they will be solved with crdt changes

ravinderjangra:
I think, we will have to disable the benchmarks as well since I have updated the `safe-nlt` dependency.

bochaco:
yeah, it seems so

bochaco:
I kept the benchmark running with your vault @ravinderjangra for now, and disabled only the fleming tests.

PR #574, opened 2020-06-10
feat(vaults): run a vault and join a running section/network

ravinderjangra:


safe-cli/subcommands/vault.rs
```diff
@@ -19,14 +19,28 @@ pub enum VaultSubCommands {
...
+        #[structopt(long = "vault-path")]
+        /// Path where to install safe-vault executable (default ~/.safe/vault/). The SAFE_VAULT_PATH env var can also be used to set the path
+        #[structopt(long = "vault-path", env = "SAFE_VAULT_PATH")]
+        vault_path: Option<PathBuf>,
+        /// Vebosity level for vaults logs
+        #[structopt(short = "y", parse(from_occurrences))]
+        verbosity: u8,
+        /// Hardcoded contacts (endpoints) to be used to bootstrap to an already running network.
+        #[structopt(short = "h", long = "hcc")]
+        hard_coded_contacts: Option<String>,
```
lionel1704, line 38: Can be a HashSet of SocketAddr or something similar

ravinderjangra: I'm not sure if we need it to be that, we can just keep it as a simple string which anyway will be parsed on the vault side.

lionel1704: The hash set is parsed automatically by structopt. So it will be simpler / less maintenance. I'm not particular about this though. It's find with me.

ravinderjangra: actually I think, even if I change this then anyway I'll have to convert this to string to pass it to nlt and same thing there as well

PR #580, opened 2020-06-17
feat(api): Implements support for storing, retrieving and resolving symlinks

dan-da:
For a detailed description of these changes, see the [README-symlinks.md](https://github.com/maidsafe/safe-api/blob/7ef8372d949c19e711b7a7429338b509644cee1a/safe-cli/README-symlinks.md) included in this PR.

**Highlights**

API:
* adds symlink_target and symlink_target_type metadata fields to FileItem
* adds follow_links flag to PUT related public APIs
* implements FilesMap::realpath() trait/func to resolve paths containing ./ ../ and relative symlinks
* modifies Safe::fetch() and Safe::inspect() APIs to resolve URL paths via FilesMap::realpath()
* modifies SafeUrl (XorUrlEncoder) to parse url path differently, because rust-url normalizes '../' away.

FFI:
* adds follow_links flag to PUT related public APIs

CLI:
* modifies 'files ls' and 'files tree' to display symlink -> target in output
* modifies 'files get' to create symlinks in local filesystem

Test:
* adds integration tests for absolute and relative symlinks
* adds test_symlinks directory tree as input for relative symlink tests
* improves integration test common/utility funcs

Doc:
* includes README-symlinks.md with a user-guide plus technical details regarding symlinks
* updates README.md with a small section about symlink support.

**new test_symlinks directory**

I created a new directory safe-api/test_symlinks for testing symlinks.  It could have been placed beneath testdata dir, but that would have required updating a lot of existing tests to anticipate correct file count, tree layout, etc.

**Windows Tests**

As of this writing, the symlink tests pass on Windows, but only if the shell running the build (including git clone) is invoked with "Run as Administrator" or running Windows 10+ with *Developer Mode* enabled.  (I have not actually performed the latter).

<strike>I plan to update this PR soon with modifications to the symlinks tests so that they will also pass if run without permissions to create symlinks.</strike>  Edit: Done.


dan-da:
These failures in the [windows-latest cli-mock-tests](https://github.com/maidsafe/safe-api/pull/580/checks?check_run_id=782079061) are almost certainly due to lacking permissions.

I will update the tests to not fail in this case, but it would probably be better if the CI harness can [run with the elevated perms](https://github.com/maidsafe/safe-api/blob/a286eae9af5d896acc31f02a5ade6c5d7c333a36/safe-cli/README-symlinks.md#symlinks-and-windows) to test the symlink functionality.  Best would be to test both conditions, but that may be asking a bit much.  :-)

```
failures:
    calling_cat_symlinks_resolve_dir_and_file
    calling_cat_symlinks_resolve_dir_outside
    calling_cat_symlinks_resolve_infinite_loop
    calling_cat_symlinks_resolve_parent_dir
```

dan-da:
Ok, I made an update so that test cases on windows for symlinks check:
* if pre-existing symlinks (in test dir) are valid  ( for read tests )
* if symlinks can be written (for write tests)

If the necessary condition is true, then the test runs.  Otherwise, it simply returns Ok.   This allows the tests to run successfully even when symlink perms are not granted.

So the next CI run for windows *should* pass, fingers crossed.

PR #586, opened 2020-06-30
feat/ffi: expose sequence data APIs from the ffi

ravinderjangra:

### DIFF
--- a/safe-ffi/ffi/fetch.rs
+++ b/safe-ffi/ffi/fetch.rs
@@ -12,7 +12,7 @@ use super::{
     errors::{Error, Result},
     ffi_structs::{
         files_map_into_repr_c, wallet_spendable_balances_into_repr_c, FilesContainer,
-        NrsMapContainer, PublicImmutableData, SafeKey, Wallet,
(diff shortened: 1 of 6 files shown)


safe-ffi/ffi/fetch.rs
```diff
@@ -29,14 +29,15 @@ pub unsafe extern "C" fn fetch(
     user_data: *mut c_void,
     start: u64,
     end: u64,
-    o_published: extern "C" fn(user_data: *mut c_void, data: *const PublicImmutableData),
-    o_wallet: extern "C" fn(user_data: *mut c_void, data: *const Wallet),
-    o_keys: extern "C" fn(user_data: *mut c_void, data: *const SafeKey),
-    o_container: extern "C" fn(user_data: *mut c_void, data: *const FilesContainer),
-    o_nrs_map_container: extern "C" fn(user_data: *mut c_void, data: *const NrsMapContainer),
-    o_err: extern "C" fn(user_data: *mut c_void, result: *const FfiResult),
+    published_cb: extern "C" fn(user_data: *mut c_void, data: *const PublicImmutableData),
```
bochaco, line 32: Perhaps needs to be renamed to `public_immd_cb` or something like that instead of `published` ?

PR #590, opened 2020-07-01
chore(ci): set git to checkout symlinks on Windows

S-Coyle:
Resolves an issue with tests failing on Windows due to symlinks not being checked out properly.


S-Coyle:
Note that the CI failure in this PR is a separate known issue, it can be ignored for the purposes of this PR

Issue #436, opened 2020-03-04
enhancement: filter files by wildcard, range, sets, similar to bash.


dan-da:
This would apply to any `safe files` operation that deals with selecting files, such as `safe files ls` and `safe files cat` and in the future perhaps `safe files copy`, `safe files move`.

In bash we can do some neat things, eg:
```
$ echo testdata/*.md
testdata/another.md testdata/test.md

$ echo photo{2,3,4,5}.jpg
photo2.jpg photo3.jpg photo4.jpg photo5.jpg

echo photo{2..5}.jpg
photo2.jpg photo3.jpg photo4.jpg photo5.jpg
```

It would be great if we can support filtering this way.

There are two aspects to this.

1. interpreting the wildcards ourselves.  eg, for `safe files ls <url>/*.jpg`, this is a remote network object, so *.jpg cannot be expanded locally.   Ideally, this would be done server-side to minimize network traffic.

2. accept multiple arguments for local expansion.  eg, if the command is `safe files ls <url>/readme{1..3}.md`, then this locally (in bash) expands and `safe files ls` receives 3 args, so it should be able to handle that, just as unix 'ls' can.

PR #605, opened 2020-07-07
feat(api): implement glob() support for pattern matching via SafeUrl.

dan-da:
This patch addresses #436 (aspect 1).

This is an experimental adaptation of [rust-glob](https://docs.rs/glob/0.3.0/glob/), embedded in Safe::fetch() URL resolution, so it affects the output of `cat`, `files ls`, `files get`, etc.

Basically, the rust-glob code has been adapted to use a Safe FilesMap instead of the native filesystem.  Also, Safe::fetch() has been modified to process output of glob() in a loop, instead of expecting only a single result.

This makes it possible to list or retrieve files using the patterns supported by glob(), eg *.jpg, photo[0-9].jpg, etc.  Even arbitrary depth recursion is supported via /**/*.jpg.  See [pattern docs](https://docs.rs/glob/0.3.0/glob/struct.Pattern.html).

I call it experimental because there are some unresolved issues.  See below.  I actually have mixed feelings if we should merge the code in its present state or not.

On the plus side, it is very nice to be able to list or get a subset of files in a directory (or even recursively) using wildcards/patterns.  This makes the safe-cli feel more flexible and familiar for those accustomed to with CLI interfaces/shells, and I think has real utility.  A couple examples:

```
$ safe cat safe://hnyynyxp8tstb5jn4i1r8uapco6onm4n91y35zx67fa9i5fzjp4dfjzrywbnc/sub/r*.md
Files of FilesContainer (version 0) at "safe://hnyynyxp8tstb5jn4i1r8uapco6onm4n91y35zx67fa9i5fzjp4dfjzrywbnc/sub/r*.md":                              
+----------------+---------------+------+----------------------+----------------------+-------------------------------------------------------------------+             
| Name           | Type          | Size | Created              | Modified             | Link                                                              |             
+----------------+---------------+------+----------------------+----------------------+-------------------------------------------------------------------+             
| /sub/readme.md | text/markdown | 34   | 2020-07-07T07:19:54Z | 2020-07-07T07:19:54Z | safe://hbhynynt39cagufa6w3sawrhm4j881ms4kc5xbz7w9zan7iqc5oeex9fx4 |             
+----------------+---------------+------+----------------------+----------------------+-------------------------------------------------------------------+ 
```

```
$ safe cat safe://hnyynyxp8tstb5jn4i1r8uapco6onm4n91y35zx67fa9i5fzjp4dfjzrywbnc/[abr]*
Files of FilesContainer (version 0) at "safe://hnyynyxp8tstb5jn4i1r8uapco6onm4n91y35zx67fa9i5fzjp4dfjzrywbnc/[abr]*":
+----------------------+---------------+------+----------------------+----------------------+-------------------------------------------------------------------+
| Name                 | Type          | Size | Created              | Modified             | Link                                                              |
+----------------------+---------------+------+----------------------+----------------------+-------------------------------------------------------------------+
| /absolute_links.txt  | text/plain    | 391  | 2020-07-07T07:19:54Z | 2020-07-07T07:19:54Z | safe://hbhydynqn7x7nnetr4zes11p5m5uw3przixkxcj9no8irum5sg6duztdp8 |
+----------------------+---------------+------+----------------------+----------------------+-------------------------------------------------------------------+
| /broken_rel_link.txt | inode/symlink | 0    | 2020-07-07T07:19:54Z | 2020-07-07T07:19:54Z |                                                                   |
+----------------------+---------------+------+----------------------+----------------------+-------------------------------------------------------------------+
| /realfile.txt        | text/plain    | 21   | 2020-07-07T07:19:54Z | 2020-07-07T07:19:54Z | safe://hbhydynbob5me99crkfqw3g1hsdqq9j9tf7qapj5xoctciaec68izzr97n |
+----------------------+---------------+------+----------------------+----------------------+-------------------------------------------------------------------+
```

In the minus column:

1. This implementation is very inefficient. glob() reads each directory component in the path.  Our FilesMap implementation must loop through all paths in the entire FilesMap to filter only the direct children of a particular directory.  In a worst case scenario of `**` recursion, this loop can occur for every directory in the FilesMap.  Obviously, this means performance for large FilesMap will suffer compared to prior Safe::fetch() implementation.   If we were to re-implement FilesMap as a tree structure, this limitation would disappear.   ( note: I made a quick prototype of such structure in a dynamic language that could be ported if we want to go that way. )

2. There is a tension/complexity in Safe::fetch() implementation, because the pre-existing code there strips out director(ies) that are present in input path in the results.  But glob() needs to return them, because the directories themselves can be part of the pattern/match.  This makes the output of `files ls` inconsistent depending if path is a pattern or not.  I think if we decide to go forward with this, it may be best to always return full FilesMap path from Safe::fetch() and let the calling code (eg `files ls`) deal with stripping prefix if desired.

3. Returning only pattern matches to safe-cli can be a bit confusing for the existing sub-commands.   One example: `safe files tree <url>/**/*.md` will display all the .md files in the tree, but no directory entries are returned.  It actually displays pretty well, except that directories are not highlighted blue, as they normally are. `safe files ls <url>/**/*.md` displays only the top level directories.  It turns out that eg real unix ls has special display modes for this type of thing.   More work would be required to make `safe ls` do the equivalent.

4. If/when we implement a fuse type filesystem integrated at OS level, regular CLI tools would work.   ie, tools such as cp, ls, etc would  `just work` and they already have glob support built in, so we get it for free.  In other words, perhaps this patch is re-inventing the wheel, and dev time is better spent in that direction instead.

**Testing**

I included most of the existing rust-glob unit tests, and added a few more.  I also added many integration tests using `safe cat` to try out the various pattern options, and also to ensure glob() can traverse and match symlinks.  That said, it wouldn't surprise me if there are still some gotchas/issues.

**Licensing**

* The code in glob.rs is adapted from rust-glob, with the fewest changes I could possibly make.  As such, the original license text is in there.   I'm not sure what project policy is in that regard.  I could move it into a separate crate, such as safe-glob using my own github account, if that helps.  ( Though maybe that would create a circular dependency?)

### DIFF
--- a/safe-api/src/api/app/fetch.rs
+++ b/safe-api/src/api/app/fetch.rs
@@ -9,8 +9,8 @@
 
 use super::{
     files::{FileItem, FileMeta, FilesMap},
+    glob,
     nrs_map::NrsMap,
-    realpath::RealPath,
     Safe, XorName,
 };
 pub use super::{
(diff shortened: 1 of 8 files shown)


theWebalyst:
Very interesting to read this as I'm looking around in the safe sync code - actually for a different reason than glob, but it has me wondering about the structure of the code. It is already complex (some large functions with lots of parameters) and I'm wary of touching what is hard to get a full understanding of. 

I searched the `safe-api` repo for `trait` because I wondered if it may be useful to use more data structures. I wanted to see how they were being used because this might be a way to reduce that complexity. The only trait implemented in safe-api appears to be SafeApp so I'm wondering if this whole area could be refractored in this direction rather than continuing to add to the complexity. (BTW, this is not a comment on this PR which I haven't looked at).

Just idle thoughts at this point, but I want to share them while I'm learning to hear any responses, and also as first impressions may be useful. Not really on topic I admit, but I didn't want to post this on the forum.

dan-da:
@theWebalyst As I understand it, the file API functions in safe-api are intended to be quite high level, for working with a FileContainer, and they do not follow the "normal" file system tree API pattern, eg opendir(), readdir(), fopen(), fwrite(), fread(), etc.  Being high-level, these APIs make certain operations very simple/easy, but they can also make it difficult or very inefficient to achieve other goals, such as your desired bi-directional sync.  And, as you've noticed the API implementations themselves have become a bit complex as more functionality is added.   As for a major refactor, I wouldn't be opposed to that, but its not really my call, and of course that would take a lot of time, resources, etc. 

@oetyng previously created a lower-level filesystem, [Safe.NetworkDrive](https://github.com/oetyng/SAFE.NetworkDrive) that could be directly mounted in windows.  We may be able to incorporate some of that design, which should then enable locally mounting SAFE Network FileContainers and accessing them with regular system tools and filesystem APIs, including rsync.  So far, this is only an idea, nothing concrete.

PR #619, opened 2020-09-09
chore(name): update safe-api repo/crate name to sn_api

S-Coyle:
Have not changed the S3 bucket names.
Also included renaming SAFE to Safe in any documentation as I came
across it.
Cargo fmt fixes included.

Note that there are 3 clippy errors locally - these seem to be related to safe_core rather than any of my updates


bochaco (approved):
I looks ok at very high level, I think we can merge it and then start testing with `cargo publish --dry-run` and clippy...etc.

PR #620, opened 2020-09-09
update safe-cli crate name to sn_cli

S-Coyle:
Have based on branch from PR #619 

At this point I have skipped renaming the artifacts produced, these can be renamed in a future PR


bochaco:
Superseded by #625

PR #631, opened 2020-10-09
Explore updates for apis

joshuef:
- Remove mock `api`
- Removes FFI
- Updates to `sn_data_types`
- MData -> Map
- Idata -> Blob
- Other related renaming (pub->public etc)
- api renaming updates
- add auth and ipc funcitonality for `authd` and `cli`
- updates cli to api
- updates authd

## What this doesn't do:

- make tests run (though they do compile). This needs fully working sn_client first.
- update for multibase update (cant test it yet as above)
- implement authenticator functionality. ipc conns can be seen, but that's all there is at the mo. Acting upon these does not yet happen.
- fix all clippy errors
- get transfer-to-xorname working yet. (This is the cause of the few `unimplemented` additions there)

## After this:

- Start basic keygeneration and returning that to clients w/ bootsdtrap on IPC.
- [later]  storing this info on the network using standard apis.
- [depends on client] : get tests going


sn_api/src/api/app/files.rs
```diff
@@ -350,7 +350,7 @@ impl Safe {
     ///     println!("FilesMap of fetched version is: {:?}", files_map);
     /// # });
     /// ```
-    pub async fn files_container_get(&self, url: &str) -> Result<(u64, FilesMap)> {
+    pub async fn files_container_get(&mut self, url: &str) -> Result<(u64, FilesMap)> {
```
bochaco, line 353: Hopefully this doesn't need to be `&mut self` but `&self` ?

sn_api/src/api/app/files.rs
```diff
@@ -359,7 +359,7 @@ impl Safe {
 
     /// Fetch a FilesContainer from a XorUrlEncoder without performing any type of URL resolution
     pub(crate) async fn fetch_files_container(
-        &self,
+        &mut self,
```
bochaco, line 362: Hopefully this doesn't need to be `&mut self` but `&self` ?

sn_api/src/api/app/files.rs
```diff
@@ -783,26 +783,25 @@ impl Safe {
...
     ///     let data = b"Something super good";
-    ///     let xorurl = safe.files_put_public_immutable(data, None, false).await.unwrap();
-    ///     let received_data = safe.files_get_public_immutable(&xorurl, None).await.unwrap();
+    ///     let xorurl = safe.files_store_public_blob(data, None, false).await.unwrap();
+    ///     let received_data = safe.files_get_public_blob(&xorurl, None).await.unwrap();
     ///     assert_eq!(received_data, data);
     /// # });
     /// ```
-    pub async fn files_get_public_immutable(&self, url: &str, range: Range) -> Result<Vec<u8>> {
+    pub async fn files_get_public_blob(&mut self, url: &str, range: Range) -> Result<Vec<u8>> {
```
bochaco, line 791: Hopefully this doesn't need to be `&mut self` but `&self` ?

sn_api/src/api/app/files.rs
```diff
@@ -783,26 +783,25 @@ impl Safe {
...
-            .await
+        self.fetch_public_blob(&xorurl_encoder, range).await
     }
 
-    /// Fetch an ImmutableData from a XorUrlEncoder without performing any type of URL resolution
-    pub(crate) async fn fetch_public_immutable_data(
-        &self,
+    /// Fetch an Blob from a XorUrlEncoder without performing any type of URL resolution
+    pub(crate) async fn fetch_public_blob(
+        &mut self,
```
bochaco, line 799: Hopefully this doesn't need to be `&mut self` but `&self` ?

sn_api/src/api/app/nrs.rs
```diff
@@ -38,7 +38,7 @@ impl Safe {
     // this second XorUrlEncoder instance contains the information of the parsed NRS-URL.
     // *Note* this is not part of the public API, but an internal helper function used by API impl.
     pub(crate) async fn parse_and_resolve_url(
-        &self,
+        &mut self,
```
bochaco, line 41: Hopefully this doesn't need to be `&mut self` but `&self` ?

sn_api/src/api/app/nrs.rs
```diff
@@ -233,27 +233,27 @@ impl Safe {
...
-    ///     let file_xorurl = safe.files_put_public_immutable(&vec![], Some("text/plain"), false).await.unwrap();
+    ///     let file_xorurl = safe.files_store_public_blob(&vec![], Some("text/plain"), false).await.unwrap();
     ///     let (xorurl, _processed_entries, _nrs_map) = safe.nrs_map_container_create(&rand_string, &file_xorurl, true, false, false).await.unwrap();
     ///     let (version, nrs_map_container) = safe.nrs_map_container_get(&xorurl).await.unwrap();
     ///     assert_eq!(version, 0);
     ///     assert_eq!(nrs_map_container.get_default_link().unwrap(), file_xorurl);
     /// # });
     /// ```
-    pub async fn nrs_map_container_get(&self, url: &str) -> Result<(u64, NrsMap)> {
+    pub async fn nrs_map_container_get(&mut self, url: &str) -> Result<(u64, NrsMap)> {
```
bochaco, line 243: Same comment about `&mut self`a before...?

sn_api/src/api/app/sequence.rs
```diff
@@ -65,7 +65,7 @@ impl Safe {
     ///     assert_eq!(received_data, (0, data.to_vec()));
     /// # });
     /// ```
-    pub async fn sequence_get(&self, url: &str) -> Result<(u64, Vec<u8>)> {
+    pub async fn sequence_get(&mut self, url: &str) -> Result<(u64, Vec<u8>)> {
```
bochaco, line 68: Same comment about `&mut self`a before...?

sn_api/src/api/app/sequence.rs
```diff
@@ -74,15 +74,15 @@ impl Safe {
 
     /// Fetch a Sequence from a XorUrlEncoder without performing any type of URL resolution
     pub(crate) async fn fetch_sequence(
-        &self,
+        &mut self,
```
bochaco, line 77: Same comment about `&mut self`a before...?

sn_api/src/api/app/wallet.rs
```diff
@@ -221,13 +224,13 @@ impl Safe {
     }
 
     pub async fn wallet_get_default_balance(
-        &self,
+        &mut self,
```
bochaco, line 227: Same comment about `&mut self`a before...?

sn_api/src/api/app/files.rs
joshuef: It does at the mo, due to `sn_client` needing to be `mut` to update `conneciton_manager` I think

bochaco: We can tackle this separately then, as discussed offline we need to look at the option of interior mutability if we are happy with it.

bochaco (approved):

PR #634, opened 2020-10-20
Chore generalised keys

joshuef:


sn_api/src/api/app/keys.rs
```diff
@@ -8,93 +8,96 @@
...
-    pub sk: String,
-}
-
 impl Safe {
     // Generate a key pair without creating and/or storing a SafeKey on the network
-    pub fn keypair(&self) -> Result<BlsKeyPair> {
-        let key_pair = KeyPair::random();
-        let (pk, sk) = key_pair.to_hex_key_pair()?;
-        Ok(BlsKeyPair { pk, sk })
+    pub fn generate_random_keypair(&self) -> Result<Keypair> {
```
Yoga07, line 26: Would it be better to rename them to `generate_random_ed_keypair`?

sn_api/src/api/app/keys.rs
```diff
@@ -8,93 +8,96 @@
...
+        let mut rng = OsRng;
+        let keypair = Keypair::new_ed25519(&mut rng);
+        Ok(keypair)
     }
 
     // Create a SafeKey on the network, preloaded from another key and return its XOR-URL.
-    pub async fn keys_create_and_preload_from_sk(
+    pub async fn keys_create_and_preload_from_sk_string(
```
Yoga07, line 33: Is this flow valid still? This is equivalent to the `CreateLoginPacketFor` RPC we had back then no?

sn_api/src/api/app/keys.rs
```diff
@@ -103,18 +106,40 @@ impl Safe {
...
-        let secret_key: SecretKey = sk_from_hex(sk)
-            .map_err(|_| Error::InvalidInput("Invalid secret key provided".to_string()))?;
+    // Check that the XOR/NRS-URL corresponds to the public key derived from the provided client id
+    pub async fn validate_sk_for_url(
+        &mut self,
+        secret_key: Arc<SecretKey>,
+        url: &str,
+    ) -> Result<String> {
+        let secret_key = Arc::as_ref(&secret_key);
+        let keypair = match secret_key {
```
Yoga07, line 125: Would it be better to write this match in a separate function. We seem to use this same piece of code multiple times.

sn_api/src/api/app/keys.rs
joshuef: I think so. You'd want to pay for the data from someone else.

sn_api/src/api/app/keys.rs
joshuef: hmmmmm, not sure. It's the default one... which is ed. I _think_ i'd only name it differently for other cases, perhaps. Or perhaps `generate_default_random_keypair` ?

sn_api/src/api/app/keys.rs
joshuef: atm it's used twice. I'd not normally bother with a DRY thing here until it's used another time. I can though if you think that's worthwhile

sn_api/src/api/app/keys.rs
Yoga07: Yea, my thoughts were to just mention the type of key. `generate_default_random_keypair` sounds good :+1:

sn_api/src/api/app/keys.rs
Yoga07: Yep agreed, twice is fine. But we'd need to be wary if we write the same piece again and write that as separate method then. Good for now :+1:

Yoga07 (approved):

PR #640, opened 2020-11-02
docs: add notes for sn_fs

joshuef:

### DIFF
--- a/sn_cli/README.md
+++ b/sn_cli/README.md
@@ -871,6 +871,12 @@ Success. TX_ID: 277748716389078887
 
 ### Files
 
+#### [ Warning: Underlying API to be deprecated ]
+
+The underlying files apis used in the CLI, and perhaps much of the CLI will be deprecated in order to use [`sn_fs`](https://github.com/maidsafe/sn_fs) at some point. This is a FUSE based filesystem which is much more comprehensive and performant. It's not yet known the impact (if any) this will have on the CLI commands. But if you're interested in a fiulesystem on SAFE, `sn_fs` is most definitely here you should be looking at the moment.
+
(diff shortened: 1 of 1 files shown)


sn_cli/README.md
```diff
@@ -871,6 +871,12 @@ Success. TX_ID: 277748716389078887
 
 ### Files
 
+#### [ Warning: Underlying API to be deprecated ]
+
+The underlying files apis used in the CLI, and perhaps much of the CLI will be deprecated in order to use [`sn_fs`](https://github.com/maidsafe/sn_fs) at some point. This is a FUSE based filesystem which is much more comprehensive and performant. It's not yet known the impact (if any) this will have on the CLI commands. But if you're interested in a fiulesystem on SAFE, `sn_fs` is most definitely here you should be looking at the moment.
```
S-Coyle, line 876: Couple of typos:
* fiulesystem = filesystem
* "...definitely here you should..." = "...definitely where you should..."

happybeing: It's it FUSE based, or POSIX based?

Ideally we have a POSIX compatible (ish) API, and separately we provide an application which uses it to enable it to be mounted as a FUSE compatible fs.

It would be good to say 'POSIX' if it's accurate. Maybe we haven't separated out the API yet, in which case sn_fs might need renaming sn_fuse?

dan-da: It hasn't been separated yet.  I think that a posix-ish API is a good goal.

joshuef: thanks for proofing and clarity there chaps! :bow: a commit coming in to those out

S-Coyle (approved):

PR #641, opened 2020-11-02
fix(api): remove repeated data storing in SafeAppClient::store_sequence()

### LINKED ISSUE #638: sotre_seq storing twice
after `client` refactor we do not need to append directly after storing, we can store the data in the intial call:

https://forum.safedev.org/t/trying-to-understand-store-sequence/2895/2

yanghuidong:
Fixes #638


joshuef:
thanks for this @yanghuidong !

Looking good but for the [clippy issue](https://github.com/maidsafe/sn_api/pull/641/checks?check_run_id=1343437945#step:6:435), if you can sort that and we get CI passing there we can get this merged in :+1: :bow:

joshuef (approved):

PR #644, opened 2020-11-23
feat(authd): store generated keypairs on the network

bochaco:


bochaco:
Superseeded by PR #645

PR #647, opened 2020-11-30
Feat enable connect with credentials

joshuef:


lionel1704 (approved):
Went through it and it looks good. I have one comment about using `serde_json::to_string` instead of base32z encoding which I believe we were using for all user facing data. 

We threw around some ideas of using QR codes etc for authorization and in the future we _miight_ make the strings user facing for other use cases (embed it in a url etc.). So maybe that can be base-32z eventually. Just a thought that hit me. Otherwise, there's no big issue that stands out.

PR #657, opened 2020-12-21
feat:check for existing balance + err when creating acct

joshuef:

### DIFF
--- a/sn_api/src/api/authenticator/mod.rs
+++ b/sn_api/src/api/authenticator/mod.rs
@@ -188,6 +188,17 @@ impl SafeAuthenticator {
 
         let mut client = Client::new(Some(keypair), None).await?;
         trace!("Client instantiated properly!");
+
+        // check if client data already exists
+        // TODO: Use a more reliable test for existing data...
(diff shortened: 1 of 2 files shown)


sn_api/src/api/authenticator/mod.rs
```diff
@@ -188,6 +188,17 @@ impl SafeAuthenticator {
...
         let mut client = Client::new(Some(keypair), None).await?;
         trace!("Client instantiated properly!");
+
+        // check if client data already exists
+        // TODO: Use a more reliable test for existing data...
+        let existing_balance = client.get_balance().await?;
+
+        if existing_balance != Money::from_nano(0) {
+            return Err(Error::AuthenticatorError(
+                "Client already data exists".to_string(),
```
S-Coyle, line 198: Should this say "Client data already exists"?

joshuef:
```suggestion
                "Client data already exists".to_string(),
```

S-Coyle (approved):

PR #667, opened 2020-12-28
feat(api): support transfers to Ed25519 PublicKeys

bochaco:


sn_api/src/api/app/helpers.rs
```diff
@@ -35,14 +35,12 @@ pub fn pk_to_hex(pk: &PublicKey) -> String {
...
-//     let pk_bytes = parse_hex(&hex_str);
-//     let mut pk_bytes_array: [u8; PK_SIZE] = [0; PK_SIZE];
-//     pk_bytes_array.copy_from_slice(&pk_bytes[..PK_SIZE]);
-//     PublicKey::from_bytes(pk_bytes_array)
-//         .map_err(|_| Error::InvalidInput("Invalid public key bytes".to_string()))
-// }
+pub fn pk_from_hex(hex_str: &str) -> Result<PublicKey> {
+    let pk_bytes = parse_hex(&hex_str);
+    let ed25519_pk = ed25519_dalek::PublicKey::from_bytes(&pk_bytes)
+        .map_err(|_| Error::InvalidInput(format!("Invalid public key bytes: {}", hex_str)))?;
```
joshuef, line 41:
```suggestion
        .map_err(|_| Error::InvalidInput(format!("Invalid Ed25518 public key bytes: {}", hex_str)))?;
```

joshuef: I think we should specify key type here. ED keys are only defaults, so could also be BLS.

sn_cli/README.md
```diff
@@ -843,10 +843,10 @@ Spendable balance inserted with name 'my-default-balance' in Wallet located at "
 
 #### Wallet Transfer
 
-Once a `Wallet` contains some spendable balance/s, we can transfer `--from` a `Wallet` an `<amount>` of safecoins `--to` another `Wallet` or `SafeKey`. The destination `Wallet`/`SafeKey` can be passed as an argument with `--to`, or it will be read from `stdin`.
+Once a `Wallet` contains some spendable balance/s, we can transfer `--from` a `Wallet` an `<amount>` of safecoins `--to` another `Wallet`, `SafeKey`, or public key. The destination `Wallet`/`SafeKey`/public key can be passed as an argument with `--to`, or it will be read from `stdin`.
```
joshuef, line 846:
```suggestion
Once a `Wallet` contains some spendable balance/s, we can transfer `--from` a `Wallet` an `<amount>` of safecoins `--to` another `Wallet`, `SafeKey`, or public key. The destination `Wallet`/`SafeKey`/public key can be passed as an argument with `--to`, or it will be read from `stdin`. Currently only Ed25519 PublicKeys are supported.
```

joshuef (changes requested):
minor doc sugs to clarify ed key hex only

joshuef (approved):

PR #668, opened 2021-01-01
Control self_update by a cargo feature

Thierry61:
Normally nothing changes for Maidsafe build process because self-update is an opt-out feature.

The procedure is a little more complex for user wanting to disable self-update feature:
```bash
cargo build -p sn_api --release
cd sn_cli/
cargo build --release --no-default-features --features "simulated-payouts"
cd ../sn_authd/
cargo build --release --no-default-features --features "simulated-payouts"
cd ..
```
Another possibility is to edit Cargo.toml file in sn_authd and sn_cli directories and remove self-update from default feature list:
```diff
sn_authd/Cargo.toml:

[features]
-default = ["simulated-payouts", "self-update"]
+default = ["simulated-payouts"]
simulated-payouts = ["sn_api/simulated-payouts"]
self-update = ["self_update"]

sn_cli/Cargo.toml:

[features]
-default = ["simulated-payouts", "self-update"]
+default = ["simulated-payouts"]
simulated-payouts = ["sn_api/simulated-payouts"]
self-update = ["reqwest", "self_update"]
```
And then just launch `cargo build --release`.

Disabling this feature is useful for user wanting to control the binaries they use, with no risk of being overwritten. Possible use cases:
- they want be sure of their integrity
- they have modified the code
- they have built a target not provided by Maidsafe

Additionally when self-update is disabled the crate doesn’t depend on openssl anymore which provides the following benefits:
- code can be compiled to MUSL targets
- major gain in executable size:

|  | With self-update | Without self-update | Gain | MUSL version | Gain |
|---|--:|--:|--:|--:|--:|
| safe | 31021920 | 22771952 | 27% | 21825984 | 30% |
| sn_authd | 21318280 | 16586400 | 22% | 15902848 | 25% |

Note also that I have removed macro_use of self_update in all cases, because I find ` env!("CARGO_PKG_VERSION"))` more explicit than `cargo_crate_version!()`.

And lastly, I did the same kind of modifications on GHA workflows which were accepted in qp2p and sn_node:
- Execute jobs only from Maidsafe repository
- Correct indentation in pr.yml

Another thing that is not friendly to user forking the repo is the pre-push hook (in .cargo-husk). It seems redundant with pr.yml and I let Maidsafe team think about removing it.

### DIFF
--- a/sn_cli/operations/node.rs
+++ b/sn_cli/operations/node.rs
@@ -7,6 +7,7 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
+#[cfg(feature = "self-update")]
 use super::helpers::download_from_s3_and_install_bin;
 use log::debug;
 use rand::distributions::Alphanumeric;
@@ -57,6 +58,12 @@ fn run_safe_cmd(
     Ok(())
 }
 
+#[cfg(not(feature = "self-update"))]
+pub fn node_install(_vault_path: Option<PathBuf>) -> Result<(), String> {
+    Err("Self updates are disabled".to_string())
+}
+
+#[cfg(feature = "self-update")]
 pub fn node_install(node_path: Option<PathBuf>) -> Result<(), String> {
     let target_path = get_node_bin_path(node_path)?;
     let _ = download_from_s3_and_install_bin(
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -12,7 +12,7 @@ env:
 
 jobs:
   build:
-    if: ${{ github.repository_owner == 'maidsafe' }}
+    # if: ${{ github.repository_owner == 'maidsafe' }}
     name: Build for benchmarking
     runs-on: ubuntu-latest
     if: false
@@ -40,7 +40,7 @@ jobs:
 
 
   benchmark:
-    if: ${{ github.repository_owner == 'maidsafe' }}
+    # if: ${{ github.repository_owner == 'maidsafe' }}
     name: Benchmarking ${{ matrix.benchmark }}
     needs: build
     runs-on: ubuntu-latest
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -12,6 +12,7 @@ env:
 
 jobs:
   build:
+    if: ${{ github.repository_owner == 'maidsafe' }}
     name: Build for benchmarking
     runs-on: ubuntu-latest
     if: false
@@ -39,6 +40,7 @@ jobs:
 
 
   benchmark:
+    if: ${{ github.repository_owner == 'maidsafe' }}
     name: Benchmarking ${{ matrix.benchmark }}
     needs: build
     runs-on: ubuntu-latest
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -12,6 +12,7 @@ env:
 
 jobs:
   build-cli-authd:
+    if: ${{ github.repository_owner == 'maidsafe' }}
     name: Build Component
     runs-on: ${{ matrix.os }}
     strategy:
@@ -67,6 +68,7 @@ jobs:
 
   
   build-mac-cli-authd:
+    if: ${{ github.repository_owner == 'maidsafe' }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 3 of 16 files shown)


bochaco:
Thanks @Thierry61 , even that I was thinking we could solve the musl issue by fixing it in self_update crate, it sounds like this could be a step forward at least, even if we solve it later one having a way to already generate musl builds (with no self-update) should be good. 

Can you please rebase it and resolve the conflicts and I'll review it in the next few days?

bochaco:
Actually, @Thierry61 , how did you build it for musl?, I'm still getting errors with `ring` with your branch when building it with `cargo build --release --target x86_64-unknown-linux-musl`

Thierry61:
I have rebased my branch. Note that I have chosen to allow locally `clippy::unusual_byte_groupings` to keep 3 bits groups to represent permissions, which seem clearer to me.

As self-update is opt-out it must be removed from the default feature list in sn_authd/Cargo.toml and sn_cli/Cargo.toml files. Also I don’t use cargo but [cross]( https://github.com/rust-embedded/cross) for cross compilation because I haven’t installed the necessary stuff to do it:
```bash
sed -i -e 's/default = \["simulated-payouts", "self-update"\]/default = ["simulated-payouts"]/' sn_authd/Cargo.toml sn_cli/Cargo.toml
cross build --release --target x86_64-unknown-linux-musl
```

These commands successfully build safe and sn_authd for musl target:
```bash
ls -l $(find target/x86_64-unknown-linux-musl/release/ -maxdepth 1 -executable -type f)
-rwxr-xr-x 2 ubuntu0 ubuntu0 21884088 janv.  6 21:56 target/x86_64-unknown-linux-musl/release/safe
-rwxr-xr-x 2 ubuntu0 ubuntu0 15855072 janv.  6 21:55 target/x86_64-unknown-linux-musl/release/sn_authd
```

But please note that safe doesn’t work anymore, neither with musl, nor without it. Now I get the following error:
```bash
./safe auth create --test-coins
Passphrase:
Password:
Sending request to authd to create a Safe...
[2021-01-06T22:23:56Z ERROR safe] sn_cli error: [Error] AuthdError - [Error] AuthenticatorError - Failed to store Safe on a Map: Data error -> Unexpected error: Cannot send zero-value transfers
```

5 days ago both versions were working fine.

bochaco:
> As self-update is opt-out it must be removed from the default feature list in sn_authd/Cargo.toml and sn_cli/Cargo.toml files

I think regardless if we do this the default or not, we need this feature in our builds, so I'm trying to get a musl build for CLI and authd which have the self-update feature enabled. This feature is very important at the moment for community testing.

> These commands successfully build safe and sn_authd for musl target

Thanks a lot for all the info, I'll continue with this tomorrow and try with `cross`, it seems was simpler than the way we do it for sn-node and docker. The main problem I am facing now is that by default I'm getting dynamically linked binaries, I think we want statically linked binary for CLI and authd.

bochaco:
I just gave it a quick try to `cross` and it seems to generate an statically linked binary, so it's quite promising, I'll be trying to get that into our CI to get next release with musl.

S-Coyle:
@Thierry61 Just to check, is the outcome here that self-update would **not** be on by default if these changes were merged?
(Sorry, I'm not a dev & have limited understanding of Rust, so I may be completely wrong!)

Thierry61:
No, the outcome of this PR in its current state is an opt-out self-update feature, meaning that it is activated by default.

To reverse this and make it opt-in, one line must to be changed in 2 files (sn_authd/Cargo.toml and sn_cli/Cargo.toml) as outlined in the diff of the original comment.

sn_cli/operations/node.rs
```diff
@@ -57,6 +58,12 @@ fn run_safe_cmd(
     Ok(())
 }
 
+#[cfg(not(feature = "self-update"))]
+pub fn node_install(_vault_path: Option<PathBuf>) -> Result<(), String> {
```
bochaco, line 62: I don't think we need to disable neither node install nor authd install when self-update is disabled?

bochaco (commented):
It looks ok, although to be honest, I really prefer to not have any feature as the code becomes really complicated to follow once you start having this conditionals based on features. We've had some features before and we removed them all. But I guess we can have this for now if we see value in it, moving forward I think we should really consider removing if possible.

bochaco (approved):

sn_cli/operations/node.rs
Thierry61: Initial installation and subsequent updates are at the same level. Accepting one but not the other wouldn’t be coherent in each of the 3 use cases I mention in my original comment.

Less importantly, we would also lose the 30% size reduction in the musl built. I didn’t test it but I suppose that install and update pull the same dependencies.

Anyway, while I was arguing I saw that you have approved the changes.

PR #686, opened 2021-01-20
refactor: makes path parameters generic for qjsonrpc Enpoint ctor

### LINKED ISSUE #685: Simplify Path Parameters for `qjsonrpc` interface
The parameter for `cert_base_path` in `qjsonrpc` is currently of type `&str`. If you have to do any manipulation to the path before passing it to `qjsonrpc` `Endpiont` constructors, it becomes quickly very cumbersome as it involves constructing a `Path` or a `PathBuf`, manipulating it, converting back to `&str`, then passing it back to the constructor `Endpoint::new()`. `Path` and `str` are intentionally kept separate in `Rust`, thus conversion between the two is liable to fail and requires a lot of error checking.

## Prerequisites

- [x] Using [latest rust stable](https://www.rust-lang.org/)
- [x] Using an up-to-date master branch

## Expected Behavior

The constructor should accept any "path-like" object so that manipulations to `cert_base_path` can happen prior to constructing the `Endpoint`.

The ideal final product might look like:

```
 let path = PathBuf::new();
 path.push("foo");
 path.push("bar");

let e = ClientEndpoint::new(&path, idle_timeout, do_keylog);
```
## Current Behavior

Currently, as described above, you need to do some cumbersome conversion like so:
```    
let path = PathBuf::new();
path.push("foo");
path.push("bar");

let path_str = path.to_str()
                      .ok_or(Error("here's an extraneous point of failure"))?;
let e = ClientEndpoint::new(path_str, idle_timeout, do_keylog);
```
    
In less contrived examples this gets quite verbose, especially in a async contexts. Occasionally, I found it even makes certain conversion & path manipulations intractable since they have to be massaged back into &str at the end.

**TLDR;** I propose we change the signature to `pub fn new<P: AsRef<std::path::Path>>(cert_base_path: P, /*other params*/)`.

## Context

I’m currently writing a node management extension to `sn_node` using JSON remote procedure calls over quic as per [this thread on the forum](https://forum.safedev.org/t/question-about-current-node-cleanup-policy/2935). Trying to run the manager on a separate task leads me to case two above, which creates a few headaches.

## Your Environment

- Rust version : rustc 1.49.0
- Version or Branch used : `master`
- Operating System and version Windows
- Link to your project : <TODO>

Scorch-Dev:
Resolves #685 

Refactors the parameters to `qjsonrpc` `ClientEndpoint` and `ServerEndpoint` constructors to be more generic without changing the current interface.

### DIFF
--- a/qjsonrpc/client_endpoint.rs
+++ b/qjsonrpc/client_endpoint.rs
@@ -11,7 +11,12 @@ use super::{jsonrpc::parse_jsonrpc_response, Error, JsonRpcRequest, Result, ALPN
 use crate::utils;
 use log::debug;
 use serde::de::DeserializeOwned;
-use std::{fs, path::PathBuf, str, sync::Arc, time::Instant};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+    sync::Arc,
+    time::Instant,
+};
 use url::Url;
 
 // JSON-RPC over QUIC client endpoint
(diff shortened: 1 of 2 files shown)


bochaco (approved):
Nice!

PR #692, opened 2021-01-28
Invalidate public names containing slash char

iancoleman:
This fixes the following problem

```sh
$ safe files put /tmp/data.bin
$ safe nrs create --link <filescontainer>?v=0 my/nrs/name
# this succeeds but should not, since the name is resolved as my with path nrs/name/...
$ safe cat safe://my/nrs/name/data.bin
# sn_cli error: [Error] ContentError - No data found for path "/nrs/name/data.bin/" on the FilesContainer at "safe://<filescontainer>/nrs/name/data.bin?v=0"
```

Note that backslash throws an error as expected: "Problem parsing the URL ... invalid domain character"

Possibly this (and the existing check for the space character that comes just before it) should reside in [validate_nrs_name](https://github.com/maidsafe/sn_api/blob/2dff02dc71bc3574763906c8592d32bde64337c9/sn_api/src/api/app/nrs.rs#L301)? For now keeping all this logic together seems most sensible.

### DIFF
--- a/sn_api/src/api/app/nrs.rs
+++ b/sn_api/src/api/app/nrs.rs
@@ -299,6 +299,11 @@ impl Safe {
 }
 
 fn validate_nrs_name(name: &str) -> Result<(XorUrlEncoder, String)> {
+    // validate no slashes in name.
+    if name.find('/').is_some() {
+        let msg = "NRS name/subname cannot contain a slash".to_string();
+        return Err(Error::InvalidInput(msg));
+    }
     let sanitised_url = sanitised_url(name);
     let xorurl_encoder = Safe::parse_url(&sanitised_url)?;
     if xorurl_encoder.content_version().is_some() {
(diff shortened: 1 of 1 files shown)


iancoleman:
Moved this check into `validate_nrs_name`.

This means there are now two places that nrs name validation happens, `validate_nrs_name` and also `SafeUrlParts::parse` (which checks for spaces, empty name etc). The check for slashes can't be made alongside those other checks because `SafeUrlParts::parse` firstly calls `Url::parse` which would not catch the slashes in the nrs name (instead parsing them as the path).

`validate_nrs_name` is called for
* `nrs_map_container_add`
* `nrs_map_container_create`
* `nrs_map_container_remove`

whereas the checks done in `SafeUrlParts::parse` are more generally utilized, so I feel this is an ok separation of validation.

bochaco (approved):
Thanks a lot @iancoleman !

PR #696, opened 2021-01-28
Remove unused check for whitespace

iancoleman:
This is already done by Url::parse in host.rs#L96
https://github.com/servo/rust-url/blob/31bc0da54c659094dacb1434996f10ed1b46b5ae/url/src/host.rs#L96
which throws the error on host.rs#L113
Note the removed error message does not appear in any existing tests.


dirvine (approved):
Nice catch!

PR #699, opened 2021-01-30
Invalidate nrs names with troublesome characters

iancoleman:
This PR extends [invalid domain characters](https://github.com/servo/rust-url/blob/e1344adb1c7de76b368e75a56eae824b7ab5ca42/url/src/host.rs#L90-L110). The newly excluded characters would never have any reasonable purpose being included in any url (anywhere in the url, not just in the domain, hence the function name `validate_url_chars`; it's larger in scope than the existing function `validate_nrs_name`).

An example of what currently does work (this PR would make invalid):

```sh
safe files put /tmp/data.bin

# create nrs 'a​b' with zero width space - zwsp is not part of char::is_whitespace or char::is_control
safe nrs create --link <filescontainer>?v=0 `printf "a\xe2\x80\x8bb"`
# Output is:
# New NRS Map for "safe://a​b" created at: "<xorurl>"
# +  a​b  <filescontainer>?v=0 
# Note the 'ab' above is 3 characters, copy and paste into dev tools to confirm: "a​b".length

# can be fetched using
safe cat safe://`printf "a\xe2\x80\x8bb"`/data.bin > /tmp/downloaded.bin
```

This PR is not about [confusables](https://util.unicode.org/UnicodeJsps/confusables.jsp). None of the characters excluded by this PR are in the confusables. Confusables often do have reasonable purpose in a url, whereas the excluded characters have no reasonable purpose in a url.

This PR is not about converting some unicode string into bytes before hashing into an xorurl. The unicode string may be percent-encoded or punycode or utf-8 or utf-16 or ... but that challenge is beyond the scope of this PR.

---

**Minor Details**

Should the error returned by `validate_url_chars` be `InvalidInput` or `InvalidXorurl`? I think `InvalidInput` since it applies to nrsurl as well as xorurl.

`validate_url_chars` is just a free-floating function, should it be structured somehow more neatly?

---

The PR only needs the above explanation, but a lot of other stuff came up while working on this so I've recorded it below.

---

**Discussion of confusables**

Although this is a discussion best had on the dev forum, some initial thoughts help clarify this PR.

I feel there is no reasonable way to exclude confusables because of their common use in everyday life.

An example of a confusable via homoglyph (the o character and the omicron character look identical):

```sh
echo "top"; echo -n "top" | sha256sum
# top
# 28720365c5e7476a011e4f43ac003ee5f16247a263b9d623aa85ed311d73bf39

echo `printf "t\xce\xbfp"`; echo -n `printf "t\xce\xbfp"` | sha256sum
# tοp
# 9c0d4e4ce2c234b7c358e3458c605ecfa93d1fd2dad527eb5408a01ff065c2e8
```

An example of a confusable via combining characters (the zstroke character looks identical to z + combining bar)

```sh
echo `printf "ea\xc6\xb6y"`; echo -n `printf "ea\xc6\xb6y"` | sha256sum
# eaƶy
# 1787ba2cf90a81d02f8e429ece00c2572e82370c1a2da44a8cc23b8bfd46e414

echo `printf "ea\xcc\xb5zy"`; echo -n `printf "ea\xcc\xb5zy"` | sha256sum
# ea̵zy
# 1a394aaec45b79ea19be46fb489c098d8e2faa1c2fb25fd105523f5e950d7810
```

---

**Discussion of character encoding**

Although this is a discussion best had on the dev forum, some initial thoughts help clarify this PR.

I feel that we should not rely on percent encoding (or punycode), since it converts `safe://人` to `safe://%E4%BA%BA` or `safe://%e4%ba%ba`. Each nrs name has different bytes and therefore different xorurls. I'm very open to further discussion of this.

To demonstrate why these 'equivalent' nrsurls do not convert to identical xorurls:

```sh
# hash of U+4EBA 人
echo -n `printf "\xe4\xba\xba"` | sha256sum
# 50f5d65d57290f7542201743eb314ab3440188a934e8ee36a4b4b41933c840f0

# hash of uppercase percent-encoded U+4EBA
echo -n "%E4%BA%BA" | sha256sum
# 863c579ff0da0b81eac80933df7ef5c925a0fdd95cecee91084a5bcbcea95697

# hash of lowercase percent-encoded U+4EBA
echo -n "%e4%ba%ba" | sha256sum
# 3dd5a638073dd71610d56f9f564b5d79323246431a53949df90d49d9384f60d8
```

Note that `Url::parse` has some default behavior which would become defacto-standard for any other nrs implementation (which will be *extremely* widely used across many languages and environments). eg the domain is encoded in either percent-encoding or punycode ([source](https://github.com/servo/rust-url/blob/e1344adb1c7de76b368e75a56eae824b7ab5ca42/url/src/host.rs#L43-L47)). It seems that case sensitivity is not especially standardized, eg [are IDN domain names case-sensitive](https://stackoverflow.com/a/19747810).

To give an example of how this defacto-standard is 'secretly' happening in the code, I've extended [xorurl.rs#L192](https://github.com/maidsafe/sn_api/blob/master/sn_api/src/api/app/xorurl.rs#L192-L195):

```diff
 let parsing_url = Url::parse(&url).map_err(|parse_err| {
     let msg = format!("Problem parsing the URL \"{}\": {}", url, parse_err);
     Error::InvalidXorUrl(msg)
 })?;
+println!("HOST: {}", parsing_url.host_str().unwrap());
+println!("QUERY: {}", parsing_url.query().unwrap_or("").to_string());

// safe nrs create --link <filescontainer>?v=0"&"x=`printf "\xe4\xba\xba"` `printf "\xe4\xba\xba"`
//
// output is
//
// HOST: %E4%BA%BA
// QUERY: 
// HOST: %E4%BA%BA
// QUERY: 
// HOST: %E4%BA%BA
// QUERY: 
// HOST: %E4%BA%BA
// QUERY: 
// HOST: %E4%BA%BA
// QUERY: 
// HOST: %E4%BA%BA
// QUERY: 
// HOST: hyryyryhd6yymrty4pqthb7wmgtczytr7xustunju9ykeusxqpfqymu8hthnra
// QUERY: v=0&x=%E4%BA%BA
// HOST: %E4%BA%BA
// QUERY: 
// HOST: %E4%BA%BA
// QUERY: 
// New NRS Map for "safe://人" created at: "safe://hyryygyaatj9coetwpgbmbih8r79hdkgj1c71oo3r967ctop8by64dnr4ran7a"
// +  人  safe://hyryyryhd6yymrty4pqthb7wmgtczytr7xustunju9ykeusxqpfqymu8hthnra?v=0&x=人
//
// We input a one-character unicode nrsname 人 U+4EBA
// But the nrsname is %E4%BA%BA - this 9 character string is used for xorname derivation.
// The one-character input is:
// 3 bytes long when encoded with utf-8
// 9 bytes long when percent-encoded (case insensitive)
// 7 bytes long when converted to punycode (xn--gmq) also case insensitive
// so we should ask, which bytes should be used to derive the xorname?
// But we are not explicitly choosing the encoding, it's chosen by 'chance' in Url::parse
```

I don't have enough knowledge to have any concrete suggestion or preference just now. My intuition says to separate the nrsurl display from the nrsurl derivation. However this impacts all developers who intend to use nrs without sn_api so we probably need to have a standard specified way of displaying and parsing and using nrsurls.

Some good reading on the topic is [RFC3490](https://tools.ietf.org/html/rfc3490) - Internationalizing Domain Names in Applications (IDNA)

---

**Future direction**

This feels to me like only the start of a very far-reaching feature (nrsname + nrsurl + xorurl validation).

It might be better as a standalone rust library which can also compile to wasm so it's available for apps, and can be written very clearly so it's obvious what and how to port into other languages.

The comment in xorurl.rs#L190 is relevant here: "there is room for improvement later to standardize on a single parser."

But my thinking with this PR is better to have this now than nothing, and it's easy enough to transfer later when required.


joshuef (approved):
PR code looks great. Nothing to add there really.

You're right though, definitely seems a rabbit hole we'll need to be going down. So thanks for the bg info there 👍

bochaco (approved):

PR #700, opened 2021-01-31
Authd prioritizes --config over env vars

iancoleman:
Authd credentials currently uses environment variables for login credentials (if they're available) even if the --config option is specified.

When there's both environment variables and a config flag, the config flag should take precedence since it's the most visible and expected credential.

To reproduce the unexpected behavior:

```sh
# start the network
safe node run-baby-fleming
safe auth start

# create account as jekyll
export SAFE_AUTH_PASSPHRASE=jekyll; export SAFE_AUTH_PASSWORD=jekyll
safe auth create --test-coins

# create account as hyde using --config option
echo "{\"passphrase\":\"hyde\",\"password\":\"hyde\"}" > /tmp/hyde_creds.json
safe auth create --test-coins --config /tmp/hyde_creds.json
# this throws an error 'Client data already exists'
# but it should create a new account for hyde
```

This is much more dramatic when using `safe auth unlock` which would log in successfully with an unexpected account if both the environment variable and the --config flag are set (which is exactly what happened to me, I thought I found some buggy behavior with permissions but it was just this bug happening invisibly).


sn_cli/operations/auth_daemon.rs
```diff
@@ -325,42 +326,45 @@ fn get_safe_seed(config_file: Option<String>) -> Result<SafeSeed, String> {
         return Err("Both the passphrase (SAFE_AUTH_PASSPHRASE) and password (SAFE_AUTH_PASSWORD) environment variables must be set for creating/unlocking a Safe.".to_string());
```
bochaco, line 326: It looks very good @iancoleman , I'm just wondering if we should move this validation (on L325) and actually all the code which tries to read from env var first (i.e. from L310 to L327) to be done after we discovered the config flag was not pass?

I'm just thinking that if you have only one of the the env vars set then the command will fail even when passing the `--config` arg, in which case it should work as env var should be ignored?

bochaco: I'm merging this now and will send a PR with the suggested change.

bochaco (approved):

PR #701, opened 2021-01-31
Validate length of nrs name and subname

iancoleman:
Total length not more than 255 bytes
Each subname not more than 63 bytes
Inspiration from https://www.ietf.org/rfc/rfc1035.txt

I'm open to any numbers here but feel there should be *some* limit to nrs name length, and rfc1035 seems like as good a place to start as any.


bochaco (approved):
I think it's fine, I'm not fully sure we really need this type of limits in Safe?
Let's give it a try with this and we can see moving forward how everything evolves and what users say. Also when supporting other languages in NRS names this would mean shorter names for them as they'll have to use unicode.

iancoleman:
> I'm not fully sure we really need this type of limits in Safe?

Yes I agree, we don't strictly need it. My reasons for adding it (just to record it while it's fresh in my mind)

* many url parsers will include checks on length limits.
* it's not very people-friendly to have nrs values longer than a certain point (at what point? I can't say but 255 seems too long already).
* what are the use cases for very long nrs names? Why not use xorname directly in that case? This will allow us to find out, because those people will request a change to allow longer names.
* nrs names will presumably be displayed somewhere (eg browser address bar) so length limits might help with ui/ux expectations.
* people can 'go around' sn_api if they desperately need to use longer nrs names.

Thanks for merging.

bochaco:
Sounds good! we can always change it if we find any use case with merit.

PR #703, opened 2021-02-01
feat: adds send_checked() qjsonrpc client and JsonRpcParams trait

### LINKED ISSUE #702: Feature-Gated, Type-Safe qjsonrpc Request Trait
As per [this forum post](https://forum.safedev.org/t/remote-node-control-experiment-in-rust-using-maidsafes-qjsonrpc-crate/2940/13?u=scorch) and originally discussed in [this PR comment](https://github.com/maidsafe/sn_api/pull/690#discussion_r567868556) , it seems like it might be a good idea to provide an optionally-compiled version of `qjsonrpc`'s `send()` function, in order to provide provide better compile-time guarantees on code correctness and parameter/method-name matching.

This purely convenience and adds no new functionality.

## Prerequisites

- [x] Using [latest rust stable](https://www.rust-lang.org/)
- [x] Using an up-to-date master branch

## Current Process

Given a custom request with method name `my-method`, params of type `MyParams`, and an expected result type of `ResultType`, you currently send something like the following:

```
my_client.send::<ResultType>("my-resquest", json!(MyParams))?;
```

In general, this allows a lot of free-floating constants or strings in `send()` statements. Or, more likely, the implementer will do something like the following for each and every request type, in order to create some better compile-time type checking on the parameter types:

```
fn send_my_request_type::<ResultType>(client: &EndpointClient, params: MyParams) -> Result<ResultType>{
    client.send("my-request", json!(params))
}
```

This is exacerbated when there are multiple request types using the same parameters, since they all  have similar signatures and different name variants (e.g. `send_foo`, `send_bar`, `send_baz`).

Moreover, you can't use something like `enum` to create umbrella types, which makes it very hard to use custom request types in generic function templates.

## Suggested Enhancement

I propose implementing a trait like `StructuredRequest` or `UmbrellaRequest` or something which ensures reliable conversion to and from `JsonRpcRequest` as a convenience. This would also allow an accompanying `send_structured(req: impl StructuredRequest)` function in the `qjsonrpc` API for added type safety. 

It would also allow consumers to utilize complex types like `enum` more readily. This also consolidates all of the conversion, which eases debugging.

Example:

```
enum MyRequestType {
    Foo(FooParams),
    Bar(BarParams),
}

impl StructuredRequest for MyRequestType {
    fn to_jsonrpc_request(self) -> JsonRpcRequest {
        match self {
            Foo(params) => JsonRpcRequest::new('foo', json!(params)),
            Bar(params) => JsonRpcRequest::new('bar', json!(params)),
        }
    }
}

impl TryFrom<JsonRpcRequest> for MyRequestType {
    Error = String;
    fn try_from(req: JsonRpcRequest) -> Result<Self, Self::Error>{
        match req.method.as_ref() {
            'foo' => Ok( MyRequestType::Foo(serde_json::from_value::<BarParams>(req.params) ),
            'bar' => Ok( MyRequestType::Bar(serde_json::from_value::<BarParams>(req.params) ),
            _ => Err("Unkown method!".to_string()),
        }
        
    }

}

```

Now you've conslidated all of the conversions into one place that's easy to debug, and you can simply call:
```
let req = MyRequestType::Foo(FooParams::new());
client.send_structured(req);
```

And a server can call
```
let req = MyRequestType::try_from(jsonrpc_req)?;
match req {
    MyRequestType::Foo(params) => println!("foo").
    MyRequestType::Bar(params) => println!("bar"),
}
```

Which is in general cleaner from the perspective of the client and the server since they've frontloaded the work of handling the raw strings and don't have to worry about passing constants out to client or server implementers.

Obviously this isn't strictly necessary, in most cases. Although in some cases, not only would it come in handy, but it would make certain structures possible where they wouldn't have been before. E.g. when temlating is involved and having having one umbrella request type allows flexibility. Namely, in the `RpcDaemon` crate being developed (see link above).

E.g.

The following would not be possible without a structured response type.

```
let daemon= RpcDaemon::new::<RequestType>();
```

## Your Environment

- Rust version : `rustc stable 1.49.0`
- Version or Branch used : `master`
- Operating System and version : `Windows 10 Home`

Scorch-Dev:
Closes #702

### DIFF
--- a/qjsonrpc/Cargo.toml
+++ b/qjsonrpc/Cargo.toml
@@ -28,3 +28,7 @@ quinn = {version = "~0.6.1", features=["tls-rustls"], default-features = false}
 [dev-dependencies]
 tempfile = "3"
 tokio = { version = "~0.2.21", features = ["rt-core", "macros"] }
+
+[features]
+default = []
+typesafe-params = []
--- a/qjsonrpc/src/client_endpoint.rs
+++ b/qjsonrpc/src/client_endpoint.rs
@@ -19,6 +19,9 @@ use std::{
 };
 use url::Url;
 
+#[cfg(feature = "typesafe-params")]
+use super::JsonRpcParams;
+
 // JSON-RPC over QUIC client endpoint
 pub struct ClientEndpoint {
     config: quinn::ClientConfig,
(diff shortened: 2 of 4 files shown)


Scorch-Dev:
Closing this because I'm going to roll it into a new crate instead with `RpcDaemon`

PR #706, opened 2021-02-03
refactor(cli): don't attempt to read authd credentials from env vars if --config was passed

bochaco:
Don't even try to read Safe credentials from environment variables when the --config option is specified, since --config has priority and even if it was set but one of the env vars was set the command will fail.


bochaco:
@iancoleman this is what I was suggesting before in https://github.com/maidsafe/sn_api/pull/700, what do you think?

iancoleman:
Yes this is great, it shows the first priority is config (since it is checked first), else it uses env vars. Much nicer code style.

PR #708, opened 2021-02-05
fix: remove short option used for dry run

b-zee:
Running `cargo run` yielded the following:

```
thread 'main' panicked at 'Argument short must be unique

        -n is already in use', $HOME/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-2.33.3/src/app/parser.rs:193:13
```

Apparently the node subcommand was using `-n` (for number of nodes) as well as the global command (for dry running). I've solved it by omitting the short option in the node subcommand.


bochaco (approved):
Thanks @b-zee !

PR #720, opened 2021-02-16
chore(deps): upgrade sn_client to v0.46.12 and most of all dependencies to their latest published version

bochaco:


joshuef (approved):

joshuef:
clippy seems to be hanging. adding to windows may be an issue...

PR #722, opened 2021-02-17
refactor: use get_sequence_entry

iancoleman:
This is the reason for originally opening https://github.com/maidsafe/sn_client/pull/1328


bochaco (approved):
Thanks @iancoleman !

PR #725, opened 2021-02-22
feat: allow aliases of Safe (immutable references)

b-zee:
Tweaked a few method signatures to leave out the mutability binding.
This means there can be multiple (immutable) references, instead of
one unique reference to the Safe instance.

(This is more of a question of whether this is something we want. I do not think there is a reason to have this being mutable, but if so, then this PR is also a commitment to the future to keep the API such that the instance can be borrowed immutably and therefore there can be multiple references at the same time. This might allow for easier usage in async contexts. This PR might serve as a draft to make sure as much of the API as possible aligns with these changes I've made mainly in keys.rs.)

### DIFF
--- a/sn_api/src/api/app/fetch.rs
+++ b/sn_api/src/api/app/fetch.rs
@@ -799,7 +799,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_fetch_range_public_blob() -> Result<()> {
-        let mut safe = new_safe_instance().await?;
+        let safe = new_safe_instance().await?;
         let saved_data = b"Something super immutable";
         let size = saved_data.len();
         let xorurl = safe
@@ -1012,7 +1012,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_fetch_public_blob_with_path() -> Result<()> {
-        let mut safe = new_safe_instance().await?;
+        let safe = new_safe_instance().await?;
         let data = b"Something super immutable";
         let xorurl = safe.files_store_public_blob(data, None, false).await?;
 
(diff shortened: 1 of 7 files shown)


b-zee:
An example of what this might allow is to simultaneously execute two routines (currently results in an error):

```rust
#[tokio::test]
async fn borrow_twice() {
    let mut safe = Safe::new(None, std::time::Duration::from_secs(120));
    safe.connect(None, None, None).await.unwrap();

    let func1 = safe.keys_create_preload_test_coins("5");
    let func2 = safe.keys_create_preload_test_coins("8");
    let (_r1, _r2) = futures::join!(func1, func2);
}
```

Now results in:
```
error[E0499]: cannot borrow `safe` as mutable more than once at a time
  --> src/lib.rs:19:17
   |
18 |     let func1 = safe.keys_create_preload_test_coins("5");
   |                 ---- first mutable borrow occurs here
19 |     let func2 = safe.keys_create_preload_test_coins("5");
   |                 ^^^^ second mutable borrow occurs here
20 |     let (_r1, _r2) = futures::join!(func1, func2);
   |                                   ----- first borrow later used here
```

bochaco (dismissed):
Yes, sounds good!

b-zee:
Thanks for looking into this, @bochaco! I will mark this as a draft for now and probably slowly apply these changes to more of the API. This will be developed hand-in-hand with the Node.js addon I'm working on, so it's dependent on that.

bochaco:
I right, perfect, sounds good!

b-zee:
@bochaco, this can be merged. I still might file some more PRs in the future, but for now this should be OK.

bochaco:
I think this will be good to merge it, yes, but we need to fix the clippy issues found in CI, you can run this locally with `cargo clippy  --all-targets --all-features`

b-zee:
Sorry again, haha. Did do `clippy` but not with the flags you mentioned earlier. I will learn to do it right eventually, I can feel it.

Should be fixed.

bochaco:
haha ! no worries, I'll merge it as soon as CI passes the clippy check. Thanks!

bochaco (approved):

PR #726, opened 2021-02-22
refactor: Adds bin-version subcommand to node & auth, removes -V for subcommands

### LINKED ISSUE #611: safe vault --version shows incorrect version
The vault version is shown as an old version when using command `safe vault --version`.

```text
$  safe vault --version
safe-vault 0.15.0
$  safe vault update
Checking target-arch... x86_64-unknown-linux-musl
Checking current version... v0.25.1
Checking latest released version... v0.25.1
Update status: '0.25.1'!
$  safe vault --version
safe-vault 0.15.0
```

When I run baby fleming the log shows vault version as `v0.25.1`

```text
...
Running safe-vault v0.25.1
==========================
...
```

Scorch-Dev:
Fixes #611 

See forum thread [here](https://forum.safedev.org/t/what-can-i-do-to-help/2965) for discussion, but it doesn't "fix" the behavior as might be suggested in the git issue.

The idea is that subcommands in our binary aren't versioned explicitly, so this patch removes the `-V` option when subcommands are supplied to avoid confusion between subcommand version (which we don't really support) and the version of external binaries like `sn_node`. Then this also adds a `bin-version` subcommand to `node` and `auth` to fetch the version of the external binary, so that the semantics are clearer as well as the distinction between the `cli` version and `sn_node` and `sn_authd` versions.

### DIFF
--- a/sn_cli/subcommands/auth.rs
+++ b/sn_cli/subcommands/auth.rs
@@ -18,7 +18,7 @@ const AUTH_REQ_NOTIFS_ENDPOINT: &str = "https://localhost:33002";
 #[derive(StructOpt, Debug)]
 pub enum AuthSubCommands {
     /// Shows the version of the Authenticator binary
-    Version {
+    BinVersion {
         #[structopt(long = "authd-path")]
         /// Path of sn_authd executable (default ~/.safe/authd/)
         authd_path: Option<String>,
@@ -134,7 +134,7 @@ pub async fn auth_commander(
     safe: &mut Safe,
 ) -> Result<()> {
     match cmd {
-        Some(AuthSubCommands::Version {authd_path}) => {
+        Some(AuthSubCommands::BinVersion {authd_path}) => {
             let sn_authd = SafeAuthdClient::new(endpoint);
             authd_version(&sn_authd, authd_path)
         }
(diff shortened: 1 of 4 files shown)


sn_cli/subcommands/auth.rs
```diff
@@ -17,8 +17,10 @@ const AUTH_REQ_NOTIFS_ENDPOINT: &str = "https://localhost:33002";
 
 #[derive(StructOpt, Debug)]
 pub enum AuthSubCommands {
+    #[structopt(alias("version"))]
```
bochaco, line 20: I guess this is just for backward compatibility and to deprecate it, but I think we are sill ok with simply changing it now, thus I'd say let's just remove this line and go with `--bin-version`.

bochaco (commented):
It looks good, I gave it a try and works fine for me, I added just a small comment, what do you say?

sn_cli/subcommands/auth.rs
Scorch-Dev: Yea I tossed it in there for that reason, but I tend to agree it's probably not worth it to support backwards compatibility here. There probably aren't enough scripts out there that rely on the `version` subcommand right now anyway.

bochaco (approved):

PR #731, opened 2021-02-24
feat(jsonrpc): adds spec-defined error code constants

### LINKED ISSUE #730: Add JSON-RPC 2.0 Pre-defined Error Constants to qjsonrpc
Currently, `qjsonrpc` is implemented to support the JSON-RPC 2.0 standard. That said, there are certain error codes that are defined in the spec that are not exposed by the `qjsonrpc`. This means consumers need to redefine the same constants themselves, which isn't necessary since they are in some sense part of the implementation.

## Possible Solution

Expose three useful JSON-RPC 2.0 error codes (this excludes parse error and invalid request error, which are still handled internally by the `qjsonrpc` library).

Code | Message | Meaning
--- | --- | ---
-32601 | Method not found | The method does not exist / is not available.
-32602 | Invalid params | Invalid method parameter(s).
-32603 | Internal error | Internal JSON-RPC error (from what I gather, this is the designated, server-side fallback/misc. error)

## Context

I'm toying with a `qjsonrpc` interface for my node binary, and found that this would be useful to have. It would also make sense to port this to `sn_authd` later (although that would require a bit of heavier refactoring to the existing error code scheme in `sn_authd` so it would make sense to hold off on applying these changes to `sn_authd` until that time).

Scorch-Dev:
Closes #730 

This patch would expose several useful spec-defined error codes in the `qjsonrpc` crate.

### DIFF
--- a/qjsonrpc/src/jsonrpc.rs
+++ b/qjsonrpc/src/jsonrpc.rs
@@ -20,6 +20,15 @@ const SN_AUTHD_JSONRPC_VERSION: &str = "2.0";
 const JSONRPC_PARSE_ERROR: isize = -32700;
 const JSONRPC_INVALID_REQUEST: isize = -32600;
 
+/// Spec-defined code for method not found
+pub const JSONRPC_METHOD_NOT_FOUND: isize = -32601;
+
+/// Spec-defined code for invalid method params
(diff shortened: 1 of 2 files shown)


bochaco (approved):
Thanks @Scorch-Dev !

PR #734, opened 2021-03-01
chore(ci): only run pr ci on prs, not on push

S-Coyle:


S-Coyle:
Seems to be an issue with CI not running against this PR. Can't find a reason why, so closing PR and I'll open a new one

S-Coyle:
GitHub Actions had an issue which caused the previous CI runs to fail - re-opening to see if I can get CI running again

PR #751, opened 2021-03-22
Use traits to make API more interoperable

b-zee:
One of the things making a good API is using trait bounds to accept a broader range of inputs. One such case is accepting `std::io::Read` so that multiple types can be passed, like slices and `File`s.

Feedback is welcome, there are a lot of places that can be improved in this regard.

### DIFF
--- a/sn_api/src/api/app/safe_client.rs
+++ b/sn_api/src/api/app/safe_client.rs
@@ -404,9 +404,9 @@ impl SafeAppClient {
     }
 
     // === Sequence data operations ===
-    pub async fn store_sequence(
+    pub async fn store_sequence<R: std::io::Read>(
         &self,
-        data: &[u8],
+        mut data: R,
         name: Option<XorName>,
         tag: u64,
         _permissions: Option<String>,
@@ -426,6 +426,11 @@ impl SafeAppClient {
         // The Sequence's owner will be the client's public key
         let owner = client.public_key().await;
 
+        let mut buf = vec![];
+        data.read_to_end(&mut buf).map_err(|e| {
+            Error::FileSystemError(format!("Failed to store Private Sequence data: {:?}", e))
+        })?;
+
         // Store the Sequence on the network
         let _address = if private {
             // Set permissions for append, delete, and manage perms to this application
@@ -433,7 +438,7 @@ impl SafeAppClient {
             let _ = perms.insert(owner, SequencePrivatePermissions::new(true, true));
 
             client
-                .store_private_sequence(Some(vec![data.to_vec()]), xorname, tag, owner, perms)
+                .store_private_sequence(Some(vec![buf]), xorname, tag, owner, perms)
                 .await
                 .map_err(|e| {
                     Error::NetDataError(format!("Failed to store Private Sequence data: {:?}", e))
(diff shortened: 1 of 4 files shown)


sn_api/src/api/app/safe_client.rs
```diff
@@ -426,14 +426,19 @@ impl SafeAppClient {
         // The Sequence's owner will be the client's public key
         let owner = client.public_key().await;
 
+        let mut buf = vec![];
+        data.read_to_end(&mut buf).map_err(|e| {
```
bochaco, line 430: This looks quite inefficient ...?...specially compared to `.to_vec()` / `.clone()` ...?
I was actually considering the option of using [bytes](https://docs.rs/bytes/1.0.1/bytes/struct.Bytes.html), not fully sure, I usually prefer APIs to have input args to be native data types.

As a side note, please take into account that the Sequence API will go away soon, so I wouldn't spend too much time on it right now, just as a heads up; we are adding the Register and the Multimap APIs: https://github.com/maidsafe/sn_api/pull/775

b-zee: Thanks for looking into this, @bochaco. Using the `std::io::Read` trait is advised by [Rust's API guidelines](https://rust-lang.github.io/api-guidelines/flexibility.html#functions-minimize-assumptions-about-parameters-by-using-generics-c-generic). The `Read` trait does provide `to_vec` or `clone`. It's a very general trait that is used because it is implemented by many objects, like [slices and `File`s](https://doc.rust-lang.org/std/io/trait.Read.html#implementors) as mentioned earlier. This could apply to `sn_client` as well, but `sn_api` is kind of the end API that should be used by most developers and should be as flexible as possible.

I do not think this is necessarily inefficient, though you might be right in certain cases. [In the case of `[u8]`, a `memcpy` is generally used internally, for example](https://github.com/rust-lang/rust/blob/237eab11560e8ea9008921a46ae60518011fe00f/library/std/src/io/impls.rs#L227).

This pull request is very general in its message. I would advise using generics more, so that the API does not make too many assumptions while it's not necessary at all.

bochaco: > but `sn_api` is kind of the end API that should be used by most developers and should be as flexible as possible.

I agree 100% with this, I completely agree with the goal of this PR or what it proposes. Let me study about what you share a bit more to provide a more educated opinion as I've never looked at those traits before, but from what you say it looks like we should consider them very seriously.

b-zee: It's important to note that `to_vec` also copies the bytes to a vector. So the old line 436 might have the same performance of the change I am proposing.

It's also important to note that perhaps the real function that should be changed is [`store_public_sequence`](https://docs.rs/sn_client/0.53.0/src/sn_client/client/sequence_apis.rs.html#144-171) in `sn_client`. That function should perhaps accept the `Read` trait, although I am not sure.

I don't want to pretend my solution is perfect, because I can understand your point of view where copying the value is not performant. Perhaps we should forget about this PR for now and focus on guidelines later, I'm not sure.

bochaco: I think the PR is not that important but what you are proposing to review, I think we should review what's the best way for a public API, so I at least need to study a bit more what Rust APIs should look like to be as friendly and flexible as possible. As we move towards a stable network, more we will be focusing on the API, specially around these aspects you bring up.

b-zee:
Closing because it is probably outdated in its implementation and `sn_api` is living at `safe_network`.

PR #754, opened 2021-04-01
Version change: sn_api v0.22.0; sn_cli v0.22.0; sn_authd v0.4.0

S-Coyle:


bochaco (changes requested):
Let's wait till it's confirmed which sn_client version we need, but it seems that we will need to upgrade sn_launch_tool too, to v0.1.x.

oetyng (approved):

PR #756, opened 2021-04-07
feat: pass SecretKey by reference

b-zee:
Prevent unnecessary cloning (which SecretKey doesn't even allow) to pass SecretKey to get balance from key.

### DIFF
--- a/sn_api/src/api/app/keys.rs
+++ b/sn_api/src/api/app/keys.rs
@@ -96,7 +96,7 @@ impl Safe {
     }
 
     // Check SafeKey's balance from the network from a given SecretKey string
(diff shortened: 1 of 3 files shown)


bochaco (commented):
Agreed, sounds good to me! please just solve the CI issues and we can merge it. You shoudl be able to find them with `cargo clippy --all-targets --all-features`.

b-zee:
Thanks @bochaco! Should be fixed, excuse me for not checking properly!

bochaco (approved):
ahh...no worries at all!! trust me, I have to force push several times usually because I sometimes forget about formatting or other clippy issues, and that's what we have CI there for :) , so all good!

PR #762, opened 2021-04-13
chore(sn_auth): add support for suggestions to errors in sn_auth

Dylan-DPC:
This adds support for suggestions in error messages for sn_auth
some more suggestions can be added to the error messages later


bochaco (commented):
@Dylan-DPC we should use https://crates.io/crates/anyhow, we are using this in CLI and in all other crates tests, it allows you to do the same I think you are trying to achieve and also works with any error type that has an impl of std::error::Error, which for example should be those we get from our API or any other lib we use.

PR #776, opened 2021-04-22
feat(api): support transfers to BLS public keys

bochaco:


b-zee:
@bochaco 

Is it deliberate that the work of #756 was undone with this PR?

https://github.com/maidsafe/sn_api/pull/776/commits/815ce6f0a0c7363779886b86c880df897b520843#diff-470737e274118012235c9c30bd79c2ed65d0a790037ad4782cabcc27b6df5fd3R98

bochaco:
I remember having issues when working on it, so not necessarily, if we can have a ref in the arg we can make it back again

b-zee:
Was there anything specific that introduced the issue? I'd happily introduce it back again, because transferring ownership for the key is problematic as you can't clone it (and probably should not due to security)...

bochaco:
Yeah, go ahead, it'll be good to change it back, I think the issue was with some of the data-types APIs for creating a keypair out of it.

b-zee:
I've made a topic on the dev forum about it. I think there are more things here to take into account.

https://forum.safedev.org/t/private-key-in-ram/3011?u=bzee

I'm also curious about the opinion from @madadam (although I don't know if he's still working on the code).

PR #780, opened 2021-04-23
[WIP] Add node status subcommand

kanav99:
Adds `safe node status` command which shows if the node is running and current used up storage.

```
./safe node status
Status: Stopped
Total storage used: 0 Bytes
```

Next up is `node config` or `node storage` subcommand which can configure max storage (in the same PR).


Vort:
Will `pgrep` work for Windows?

kanav99:
Nope, need to make a branch for that, PR is still a work in progress 😄

kanav99:
Needs https://github.com/maidsafe/sn_launch_tool/pull/68

happybeing:
It's there for a way for a node to know if it is still accepted in the network? If so, a way to report that works be useful.

sn_cli/subcommands/config.rs
```diff
@@ -20,6 +20,9 @@ pub enum ConfigSubCommands {
     #[structopt(name = "remove")]
     /// Remove a config setting
     Remove(SettingRemoveCmd),
+    #[structopt(name = "set-max-capacity")]
```
bochaco, line 23: I haven't look at it thoroughly yet, but the config cmds are expected to be generic, so you'd do something like `safe config add node-max-capacity <value>`, or that kind of things, we could have a `set` subcommand instead of `add` though, so we can use that for either adding or changing a config.

kanav99: Done!

sn_cli/subcommands/config.rs
```diff
@@ -67,6 +78,9 @@ pub async fn config_commander(cmd: Option<ConfigSubCommands>) -> Result<()> {
             config.add_network(&network_name, config_location.map(NetworkInfo::ConnInfoUrl))?;
         }
         // Some(ConfigSubCommands::Add(SettingAddCmd::Contact { name, safeid })) => {}
+        Some(ConfigSubCommands::Set(SettingSetCmd::NodeMaxCapacity { capacity })) => {
+            config.set_max_capaity(capacity)?;
```
kanav99, line 82: We should print out that the changes won't be reflected until node is restarted.

PR #823, opened 2021-07-14
feat: default to read url file instead of hardcoded url

grumbach:
- reactivate some API tests for CI
- default to random credentials if none are provided (instead of an error)
- default to reading urls from "$HOME/.safe/node/node_connection_info.config" if none are provided (instead of hardcoded url)


src/app/test_helpers.rs
```diff
@@ -72,9 +91,10 @@ fn get_bootstrap_contacts() -> Result<HashSet<SocketAddr>> {
             )
         })?,
         Err(_) => {
-            // we default to this address as that's what we
-            // normally use in local test network as the genesis node address
-            vec!["127.0.0.1:12000".parse()?].into_iter().collect()
+            // read default peers from the file we normally use for peers
+            let default_peer_file =
```
bochaco, line 95: I'd make this a `const` at the top of the file, and then use it directly from within `read_default_peers_from_file` fn

grumbach: Yeah I agree, I started with a 
```rust
const DEFAULT_PEER_FILE: &str = "~/.safe/node/node_connection_info.config"
```
But there was a problem with the `~` in the path so I switched to
```rust
format!("{}/.safe/node/node_connection_info.config", var("HOME")?)
```
But this is not const'able. I could const the second part of the path though. Would this be acceptable? 
```rust
const DEFAULT_PEER_FILE_IN_HOME: &str = ".safe/node/node_connection_info.config"

// and use
format!("{}/{}", var("HOME")?, DEFAULT_PEER_FILE_IN_HOME)
```
What do you think?

bochaco: We use `dirs_next::home_dir()` (https://crates.io/crates/dirs-next) in other places as this needs to work cross platform, and I think you can then do `.push(DEFAULT_PEER_FILE_IN_HOME)`

bochaco (approved):

PR #833, opened 2021-07-30
NRS with multimaps, File Containers with Registers, sequences OUT

grumbach:
BREAKING_CHANGE

Quite a big PR
- NRS with Multimaps
- File containers with Registers
- Removed all traces of the old Sequence CRDT


src/app/files/mod.rs
```diff
@@ -142,51 +146,62 @@ impl Safe {
...
+
+        // take the 1st entry (TODO Multiple entries)
+        if entries.len() > 1 {
+            unimplemented!("Multiple file container entries not managed, this happends when 2 clients write concurrently to a file container");
         }
+        let first_entry = entries.iter().next();
+        let (version, curr_serialised_files_map);
+        if let Some((v, m)) = first_entry {
+            version = v.into();
+            curr_serialised_files_map = m.to_owned();
```
davidrusu, line 177:
```suggestion
        let (version, curr_serialised_files_map) = if let Some((v, m)) = first_entry {
            (v.into(), m.to_owned())
```

src/app/files/mod.rs
```diff
@@ -142,51 +146,62 @@ impl Safe {
...
+        } else {
+            warn!("FilesContainer found at \"{:?}\" was empty", safe_url);
+            return Ok((VersionHash::default(), FilesMap::default()));
+        }
+
+        debug!("Files map retrieved.... v{:?}", &version);
+        // TODO: use RDF format and deserialise it
+        // We first obtain the FilesMap XOR-URL from the Sequence
```
davidrusu, line 185:
```suggestion
        // We first obtain the FilesMap XOR-URL from the Register
```

src/app/files/mod.rs
```diff
@@ -483,29 +498,32 @@ impl Safe {
...
+        current_version: VersionHash,
         new_files_map: &FilesMap,
         url: &str,
         mut safe_url: SafeUrl,
         dry_run: bool,
         update_nrs: bool,
-    ) -> Result<u64> {
-        let version = if success_count == 0 {
+    ) -> Result<VersionHash> {
+        let version = if success_count == 0 || dry_run {
```
davidrusu, line 508: On dry-run, we should return the new version hash as if the append succeeded

src/app/files/mod.rs
```diff
@@ -483,29 +498,32 @@ impl Safe {
...
-            self.safe_client
-                .append_to_sequence(files_map_xorurl.as_bytes(), xorname, type_tag, false)
+            // append entry to register
+            let entry = files_map_xorurl.as_bytes().to_vec();
+            let replace = self
+                .register_read(&safe_url.to_string())
+                .await?
+                .iter()
+                .map(|(h, _)| h.to_owned())
+                .collect();
```
davidrusu, line 522: Use `current_versions` provided by the user instead of reading the most recently updated version from the network.

src/app/files/mod.rs
```diff
@@ -483,29 +498,32 @@ impl Safe {
     async fn append_version_to_files_container(
         &mut self,
         success_count: u64,
-        current_version: u64,
+        current_version: VersionHash,
```
davidrusu, line 501: Make this be a set of VersionHash's so that we can over-write multiple concurrent versions with this one new version

src/app/files/mod.rs
```diff
@@ -2189,18 +2211,19 @@ mod tests {
         assert!(v1_files_map.get(file_path3).is_none());
         assert!(v1_files_map.get(file_path4).is_none());
 
-        // let's fetch version 2 (invalid)
-        safe_url.set_content_version(Some(2));
+        // let's fetch version invalid version
```
davidrusu, line 2214:
```suggestion
        // let's fetch invalid version
```

src/app/files/mod.rs
```diff
@@ -2189,18 +2211,19 @@ mod tests {
...
+        safe_url.set_content_version(Some(VersionHash::default()));
         match safe.files_container_get(&safe_url.to_string()).await {
             Ok(_) => Err(anyhow!(
-                "Unexpectedly retrieved verion 2 of container".to_string(),
+                "Unexpectedly retrieved invalid version of container".to_string(),
             )),
             Err(Error::VersionNotFound(msg)) => {
+                let default_vh = format!("{}", VersionHash::default());
```
davidrusu, line 2221: Any reason why we `format` here rather than just store the `VersionHash::default()` in the variable and format when it's needed?

src/app/files/mod.rs
```diff
@@ -2364,7 +2389,7 @@ mod tests {
             true, // dry run
         ));
 
-        assert_eq!(version, 1);
+        // skip version hash check since hash didn't change on dry run
```
davidrusu, line 2392: Dry-run should give back the version that would have been created

src/app/nrs/mod.rs
```diff
@@ -93,36 +95,40 @@ impl Safe {
...
         }
 
-        Ok((version + 1, xorurl, processed_entries, nrs_map))
+        let nrs_map_xorurl = self.store_nrs_map(&nrs_map).await?;
+        let old_values: BTreeSet<EntryHash> = self
+            .fetch_multimap_values(&safe_url)
+            .await?
+            .iter()
+            .map(|(hash, _)| hash.to_owned())
+            .collect();
```
davidrusu, line 110: We should be using the `version` we got above as the `old_values`.

We should always aim to operate within the context that the user is working within, this would be fetching data from the network that the user had not seen yet. We could be overwriting data that the user did not intend to overwrite.

src/app/nrs/mod.rs
```diff
@@ -93,36 +95,40 @@ impl Safe {
...
+
+        Ok((new_version, xorurl, processed_entries, nrs_map))
     }
 
     /// # Create a NrsMapContainer.
     ///
     /// ## Example
     ///
-    /// ```rust,no_run
+    /// ```rust_no_run
```
davidrusu, line 125: Did you intend to remove the comma here?

src/app/nrs/mod.rs
```diff
@@ -205,35 +204,43 @@ impl Safe {
...
         }
 
-        Ok((version + 1, xorurl, processed_entries, nrs_map))
+        debug!("Removing from multimap");
+        let old_values: BTreeSet<EntryHash> = self
+            .fetch_multimap_values(&safe_url)
+            .await?
+            .iter()
+            .map(|(hash, _)| hash.to_owned())
+            .collect();
```
davidrusu, line 217: Again, only operate on data that the user knows about. Use the version hash you got above for the removal.

This could remove entries that the user had not seen yet since we are doing a network fetch here.

davidrusu (commented):
We shouldn't always use the version's provided by the user when we are operating on registers / multi-maps, otherwise you are risk making changes to data the user had not intended to change.

Also, in your tests, you use `assert!(x != y)` in a lot, use `assert_ne!(x, y)` instead.

src/app/files/mod.rs
```diff
@@ -142,51 +146,60 @@ impl Safe {
...
+                        None => "None".to_owned(),
+                    },
+                    safe_url
+                )),
+                err => Error::NetDataError(format!("Failed to get current version: {}", err)),
+            })?;
+
+        // take the 1st entry (TODO Multiple entries)
+        if entries.len() > 1 {
+            unimplemented!("Multiple file container entries not managed, this happends when 2 clients write concurrently to a file container");
```
bochaco, line 171: I think it'll be better to return an error so it's properly reported to the user explaining the limitation of this lib/API, otherwise CLI (or the app using the lib) crashes and provides a dump stack trace report which users won't understand.

src/app/files/mod.rs
grumbach: Yes indeed but since we can't obtain a new hash because no network (register) operations are done, what should we return? We can't increment the hash like we did for u64.

src/app/nrs/mod.rs
```diff
@@ -205,35 +204,43 @@ impl Safe {
...
+
+        // TODO use remove
```
davidrusu, line 219: Let's create an issue to track this TODO

src/app/files/mod.rs
grumbach: Added to TODO list when we upgrade to manage forks and DAG related conflicts

grumbach: Added a special unimplemented case

src/app/files/mod.rs
grumbach: Will add this to the next PR when we manage DAG related conflicts, forks and resolutions

src/app/files/mod.rs
grumbach: Skip that check for now (not implemented)

src/app/nrs/mod.rs
```diff
@@ -93,22 +94,23 @@ impl Safe {
...
-                .append_to_sequence(
-                    nrs_map_xorurl.as_bytes(),
-                    safe_url.xorname(),
-                    safe_url.type_tag(),
-                    false,
-                )
-                .await?;
+
+        if dry_run {
+            return Ok((version, xorurl, processed_entries, nrs_map));
```
davidrusu, line 100: This is not the new version, return Error::Unimplemented instead

src/app/nrs/mod.rs
```diff
@@ -205,35 +199,39 @@ impl Safe {
...
-            self.safe_client
-                .append_to_sequence(
-                    nrs_map_xorurl.as_bytes(),
-                    safe_url.xorname(),
-                    safe_url.type_tag(),
-                    false,
-                )
-                .await?;
+        if dry_run {
+            return Ok((version, xorurl, processed_entries, nrs_map));
```
davidrusu, line 203: This is not the correct version, return Unimplemented instead

davidrusu (commented):
Looking good, just a few more notes

src/app/nrs/mod.rs
```diff
@@ -93,22 +94,25 @@ impl Safe {
...
-                    nrs_map_xorurl.as_bytes(),
-                    safe_url.xorname(),
-                    safe_url.type_tag(),
-                    false,
-                )
-                .await?;
+
+        if dry_run {
+            return Err(Error::NotImplementedError(
+                "No dry run for nrs_map_container_add (missing version info)".to_string(),
```
joshuef, line 101: This error message confuses me. 

If we had versionn info, we'd do a dry run?

how come we don't have that?

src/app/nrs/mod.rs
```diff
@@ -242,83 +244,67 @@ impl Safe {
...
+                Error::ContentNotFound(_) => {
+                    Error::ContentNotFound(ERROR_MSG_NO_NRS_MAP_FOUND.to_string())
+                }
+                Error::VersionNotFound(msg) => Error::VersionNotFound(msg),
+                err => Error::NetDataError(format!("Failed to get current version: {}", err)),
+            })?;
+
+        // take the 1st entry (TODO Multiple entries)
+        if entries.len() > 1 {
+            return Err(Error::NotImplementedError("Multiple NRS map entries not managed, this happends when 2 clients write concurrently to a NRS map".to_string()));
```
joshuef, line 265: Let's make a real error for this as opposed to the catchall string 👍

joshuef (changes requested):

src/app/nrs/mod.rs
grumbach: The version info is missing because register operations dry_run are not implemented yet on safe_network so there is no way currently to know the next version hash (like we use to with `u64` with a `+1`)

src/app/nrs/mod.rs
grumbach: We were thinking of implementing this in the near future, thus the `NotImplementedError` to remind us this is temporary. But I could make a real error out of it if makes things clearer, should I?

src/app/nrs/mod.rs
grumbach: Should I change the message with `register operations dry_run are not implemented yet`?

joshuef:
```suggestion
                "No dry run for nrs_map_container_add. Version info cannot be determined. (Register operations need this funcitnoality implemented first.)".to_string(),
```

src/app/nrs/mod.rs
joshuef: This is more concrete than the above, I think it's worthwhile adding a proper error (which we can remove later), in case this does not happen "soon".

grumbach, line 265:
```suggestion
            return Err(Error::MultimapFork("Multiple NRS map entries not managed, this happends when 2 clients write concurrently to a NRS map".to_string()));
```

joshuef (approved):

PR #840, opened 2021-08-12
URL registers, enable multi replacement capability

grumbach:
- Use the new registers that contain `NativeUrl` instead of `[u8]`
- Enable replacing multiple old versions in a helper func [as David suggested](https://github.com/maidsafe/sn_api/pull/833#discussion_r680658804)  (without changing the API yet)


src/app/multimap.rs
```diff
@@ -19,7 +19,7 @@ pub type MultimapValue = Vec<u8>;
 pub type MultimapKeyValue = (MultimapKey, MultimapValue);
 pub type MultimapKeyValues = BTreeSet<(EntryHash, MultimapKeyValue)>;
 
-const MULTIMAP_REMOVED_MARK: &[u8] = b"";
+const MULTIMAP_REMOVED_MARK: &str = "safe://MULTIMAP_REMOVED_MARK";
```
davidrusu, line 22: I'm not sure this is safe, someone could potentially register this NRS url and then the remove mark suddenly resolves to this URL.., then again someone might want to store this Url in their multi map.

I think we need enums, or if we can restrict register entries to XorNames only (i.e. resolved urls) then we can designate 0<sup>256</sup> as the sentinel.

src/app/multimap.rs
```diff
@@ -184,9 +199,10 @@ impl Safe {
         }?;
 
         // We parse each entry in the Register as a 'MultimapKeyValue'
-        if entry == MULTIMAP_REMOVED_MARK {
+        if entry_ptr.to_string() == MULTIMAP_REMOVED_MARK {
```
davidrusu, line 202: This feels a bit sloppy, we shouldn't convert to string to check if this is a remove sentinel. Instead we should compare the entry ptr itself to the removed mark (ideally we're dealing with a xor name in which case we just compare to [0u8; 32]).

davidrusu (commented):
Looks good overall, but I think the remove mark needs some work. This might require some work on the register type in safe_networks/types/register to allow for name spacing.

src/app/multimap.rs
grumbach: Agreed

grumbach: We could remove `MULTIMAP_REMOVED_MARK` altogether as `multimap_remove` is unused by NRS for now.

connec (approved):

PR #841, opened 2021-08-12
chore: Update `xor_name`

connec:
- 23067db **chore: update `xor_name`**

  `xor_name` 2.0 has been released and a bunch of our dependencies depend
  on it. As such we're currently falling victim to shenanigans, where
  cargo is pulling in both versions and things are failing to compile.


jacderida (approved):

connec:
Fixed by #840.

PR #843, opened 2021-08-30
feat: files_container_create xorurl with version hash

grumbach:

### DIFF
--- a/src/app/files/mod.rs
+++ b/src/app/files/mod.rs
@@ -107,11 +107,11 @@ impl Safe {
             )?;
 
             let entry = NativeUrl::from_xorurl(&files_map_xorurl)?;
-            let _ = &self
+            let entry_hash = &self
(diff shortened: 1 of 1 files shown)


src/app/files/mod.rs
```diff
@@ -107,11 +107,11 @@ impl Safe {
             )?;
 
             let entry = NativeUrl::from_xorurl(&files_map_xorurl)?;
-            let _ = &self
+            let entry_hash = &self
                 .write_to_register(&xor_url, entry, Default::default())
                 .await?;
 
-            xor_url
+            format!("{}?v={}", xor_url, VersionHash::from(entry_hash))
```
bochaco, line 114: I think we should use the NativeUrl API for this, otherwise if the URLs format change (e.g. we stop using `v=<version>` for version/hash) this URL will be invalid

